PR #52, opened 2024-10-17
Fix: querybuilder updates

switzerb:
Small update for the querybuilder component to ensure classnames are merged in and passed to children. 

From this MR while it still has c2ds embedded in the application context: https://gitlab-cbc2.cce.af.mil/abms/cbc2/jeric2o/jeric2o/-/merge_requests/702

### DIFF
--- a/.changeset/gold-bees-smile.md
+++ b/.changeset/gold-bees-smile.md
@@ -1,5 +1,5 @@
 ---
-"@accelint/design-system": minor
+"@accelint/design-system": patch
 ---
 
-patch: fix small bug where classnames didn't propagate
+Fixed bug where classnames didn't propagate for query builder component.
--- a/.changeset/gold-bees-smile.md
+++ b/.changeset/gold-bees-smile.md
@@ -1,0 +1,5 @@
+---
+"@accelint/design-system": minor
+---
+
+patch: fix small bug where classnames didn't propagate
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,8 @@ dist
 # Environment variables
 .env
 
+# ide files
+*.iml
+.idea
+
+
--- a/packages/design-system/src/components/query-builder/group.tsx
+++ b/packages/design-system/src/components/query-builder/group.tsx
@@ -115,7 +115,6 @@ export function RuleGroup(props: RuleGroupProps) {
       <div
         ref={group.previewRef}
         className={group.outerClassName}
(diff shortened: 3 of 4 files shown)


brandonjpierce:
Changeset needed since this is actual package code change

.changeset/gold-bees-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/design-system": minor
```
brandonjpierce, line 2: Nit: your changelog message says `patch` but you marked this as a `minor`

.changeset/gold-bees-smile.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+patch: fix small bug where classnames didn't propagate
```
brandonjpierce, line 5: Nit: can make this more human readable / past tense e.g.

```
Fixed bug where classnames didn't propagate for query builder component.
```

.changeset/gold-bees-smile.md
switzerb: i've actually been a little confused by which language belongs with which change. so a `patch` is a fix but a feature would have a `minor` (or major) label?

ArrayKnight (dismissed):

.changeset/gold-bees-smile.md
brandonjpierce: A little more nuanced but realistically the loose definitions provided [here](https://docs.npmjs.com/about-semantic-versioning) can be used as a good rule of thumb.

brandonjpierce:
Notice a minor warning with `rollup` and our ts aliases: 

Screenshot 2024-10-17 at 2 07 13 PM

Potentially okay but figured i'd note it here. Can replicate by running `pnpm run build`

brandonjpierce:
On further investigation the alias warning may indeed have some unfortunate side effects:

https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency

switzerb:
> On further investigation the alias warning may indeed have some unfortunate side effects:
> 
> https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency

looks like maybe the aliasing isn't a good idea right now. i'll revert that bit.

brandonjpierce:
Yeah on the TODO list to redo the build config for design system a bit. It is currently the only package where we have this issue since it needs rollup.

PR #53, opened 2024-10-17
update option and menu component to include border radius

switzerb:
Updates options and menu component to include border radius style as well as updating the ladle theme to reflect the small padding gap between the top of the dropdown and the first/last item on the list, so it is more consistent with the theme as designed.

CleanShot 2024-10-17 at 11 44 40

Also changed the fallback for the y axis padding so that it will show the gap (and border-radius) unless it is specifically overridden.

### DIFF
--- a/packages/design-system/.ladle/theme/menu.css.ts
+++ b/packages/design-system/.ladle/theme/menu.css.ts
@@ -35,6 +35,9 @@ export const Menu: ThemeContext['Menu'] = {
                   x: `calc(${sizeVars.v03} + ${sizeVars.v03})`,
                   y: sizeVars.v03,
                 },
+                section: {
+                  y: sizeVars.v03
+                }
               },
             },
           ),
--- a/packages/design-system/.ladle/theme/menu.css.ts
+++ b/packages/design-system/.ladle/theme/menu.css.ts
@@ -35,9 +35,6 @@ export const Menu: ThemeContext['Menu'] = {
                   x: `calc(${sizeVars.v03} + ${sizeVars.v03})`,
                   y: sizeVars.v03,
                 },
-                section: {
-                  y: sizeVars.v03
-                }
               },
             },
           ),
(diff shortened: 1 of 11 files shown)


packages/design-system/.ladle/theme/menu.css.ts
```diff
@@ -35,6 +35,9 @@ export const Menu: ThemeContext['Menu'] = {
                   x: `calc(${sizeVars.v03} + ${sizeVars.v03})`,
                   y: sizeVars.v03,
                 },
+                section: {
+                  y: sizeVars.v03
```
ArrayKnight, line 39: Let's remove this (and the other updates to section) and use your fallback approach on the top level Menu instead

ArrayKnight (changes requested):
I realize that we have a section at the root, but I don't want to add this additional spacing to every section

packages/design-system/.ladle/theme/menu.css.ts
ArrayKnight: Since we have the fallback, it feels unnecessary to have the theme value as well

switzerb: good call. changes made.

brandonjpierce (approved):

packages/design-system/src/components/menu/menu.css.ts
```diff
@@ -118,11 +119,12 @@ export const menuClassNames: MenuClassNames = {
     }),
     list: style({
       '@layer': {
-        [layers.components.l1]: {
+        [layers.components.l2]: {
           display: 'flex',
           flexDirection: 'column',
           background: menuColorVars.list.background,
           border: `1px solid ${fallbackVar(menuColorVars.list.border, 'transparent')}`,
+          padding: `${fallbackVar(menuSpaceVars.menu.y, sizeVars.v03)} ${fallbackVar(menuSpaceVars.menu.x, '0')}`,
```
ArrayKnight, line 127: This is already on the menu / options. It shouldn't have been added to the list

ArrayKnight (approved):

brandonjpierce (approved):

PR #54, opened 2024-10-17
Fix/ds vanilla export

ArrayKnight:
Updated DS build with a quick fix (possible full rewrite to come at a later time) to include all types in the `vanilla.d.ts` export

### DIFF
--- a/.changeset/plenty-planets-act.md
+++ b/.changeset/plenty-planets-act.md
@@ -1,0 +1,5 @@
+---
+"@accelint/design-system": patch
+---
+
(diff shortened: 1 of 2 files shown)


switzerb (approved):
seems like a reasonable interim fix -- maybe add a TODO so the additional build work doesn't get lost?

brandonjpierce:
@switzerb created https://github.com/gohypergiant/standard-toolkit/issues/55

brandonjpierce (approved):

PR #57, opened 2024-10-19
fix: build packages with tsup

brandonjpierce:


packages/constants/package.json
```diff
@@ -14,14 +14,13 @@
         "types": "./dist/index.d.ts",
         "default": "./dist/index.js"
       },
-      "default": "./dist/index.js",
-      "module": "./dist/index.js"
```
brandonjpierce, line 18: Note: I yanked `module` here since `syncpack` was incorrect sorting by name, which order for `exports` does matter and so this was throwing a warning. `module` was actually redundant so no harm no foul.

packages/constants/tsup.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  entry: ['src/index.ts'],
+  dts: true,
+  format: 'esm',
+  target: 'es2023',
```
ArrayKnight, line 7: Did this end up being necessary? I assumed that it was using the same target from the tsconfig

ArrayKnight (approved):

packages/constants/tsup.config.ts
brandonjpierce: Indeed, I was made aware of this fact after making this change so this is definitely redundant

packages/constants/package.json
JamieMason: 👋 Hey @brandonjpierce, 
Would you mind [creating an issue](https://github.com/JamieMason/syncpack/issues/new?assignees=&labels=&projects=&template=bug_report.md) for this please? Syncpack is _supposed_ to [sort exports](https://jamiemason.github.io/syncpack/config/sort-exports/), so this shouldn't happen. 

I tried the following but couldn't reproduce it:

```bash
git clone git@github.com:gohypergiant/standard-toolkit.git
cd standard-toolkit
git checkout 5c41d76 # the commit before this PR
syncpack format
```

If you could let me know how, I can work on a fix – thanks!

PR #64, opened 2024-10-21
feat: adds TextField and NumberField ds components

switzerb:
Adds two new components to the design system - TextField and NumberField. These are form fields that wrap up functionality for application engineers to more easily be able to drop in common functionality like hint text and error messages.

Steps to test:

`pnpm --filter=@accelint/design-system preview` will run the ladle stories

The respective React Aria components: 

https://react-spectrum.adobe.com/react-aria/NumberField.html
https://react-spectrum.adobe.com/react-aria/TextField.html


packages/design-system/src/components/number-field/types.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type NumberFieldMapping = {
+  description: Partial<Record<NumberFieldSizes, string>>;
+  error: Partial<Record<NumberFieldSizes, string>>;
+  increment: Partial<Record<NumberFieldSizes, OmitProtectedProps<ButtonProps>>>;
+  decrement: Partial<Record<NumberFieldSizes, OmitProtectedProps<ButtonProps>>>;
+};
+
+type BaseNumberFieldProps = {
+  children?: RenderPropsChildren<NumberFieldRenderProps>;
```
ArrayKnight, line 39: You don't need to redefine `children` unless you're changing the type. This type is the same as the children type from the RAC

packages/design-system/src/components/text-field/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type TextFieldSizes = 'sm' | 'lg';
+
+export type TextFieldMapping = {
+  description: Partial<Record<TextFieldSizes, string>>;
+  error: Partial<Record<TextFieldSizes, string>>;
+};
+
+type BaseTextFieldProps = {
+  children?: RenderPropsChildren<TextFieldRenderProps>;
```
ArrayKnight, line 26: You don't need to redefine children unless you're changing the type. This type is the same as the children type from the RAC

ArrayKnight (changes requested):

ArrayKnight (approved):

PR #65, opened 2024-10-21
chore: 🤖 convert DS build to tsup, add explicit vanilla export

ArrayKnight:
Swapped Rollup out for TSup

Notes:
- I wasn't able to get the nested structure to work, so right now it's just bundling
- Due to renaming limitations in ESBuild, styles.css is now index.css, but the import in the app should be the same(?) since I updated the exports on the package.json
- Due to renaming and configuration limitations (and not wanting to write another custom plugin), the vanilla export is created from a new vanilla entry point. This allows for complete control over what is exported, but is a more manual process. Maybe there's a future way for this file to be generated based on a glob
- You'll see chunks in the `dist`. But they appear to be valid sharing of code / types between the exports without mixing anything into vanilla that we don't want


brandonjpierce:
I don't know the ramifications of this, but currently the `dist/index.js` file and `dist/vanilla.js` file are exporting the same things (with `dist/index.js` additional ones from same chunk file):

> dist/vanilla.js
Screenshot 2024-10-22 at 9 54 35 AM

> dist/index.js
Screenshot 2024-10-22 at 9 54 24 AM

brandonjpierce:
I tried doing an experiment with running three separate tsup configs one for each of our intended outputs, but was unable to get it working (or more accurately, matching the current output of the `dist` to a degree). If we wanted to hyper optimize the bundle output we may need to slightly restructure the package structure a smidge.

brandonjpierce (approved):

PR #73, opened 2024-10-23
Enable --watch with filter at top level

### LINKED ISSUE #71: Support vitest watch mode
We currently run something similar to the following in each workspace package for the `test` command:

```
pnpm run vitest --run --dir src
```

The explicit `--run` flag prevents the `--watch` flag from working as [it explicitly disables watch](https://vitest.dev/guide/cli-generated.html#run). The fix:

Update all `packages/*` to use the following for their `test` script command:

```
pnpm run vitest --dir=src
```

This way you can isolate to a package and enable watch mode for tests e.g.:

```
pnpm --filter=@accelint/math run test --watch
```

kalisjoshua:
remove --run from vitest command because it disables --watch

closes #71

### DIFF
--- a/packages/converters/package.json
+++ b/packages/converters/package.json
@@ -24,7 +24,7 @@
     "dev": "pnpm tsc --watch",
     "index": "pnpm zx ../../scripts/indexer.mjs packages/converters",
     "lint": "pnpm biome lint ./",
-    "test": "pnpm vitest --run --dir src"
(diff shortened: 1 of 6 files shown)


brandonjpierce (approved):

brandonjpierce:
Thanks for catching this! Possibly a good idea to include a quick tip in one of our documentation files on how to run watch mode for a particular package

PR #82, opened 2024-11-04
Refactor/ds lodash

ArrayKnight:


packages/design-system/src/components/button/button.tsx
```diff
@@ -1,5 +1,5 @@
 import { clsx } from 'clsx';
-import { noop } from 'lodash';
+import { noop } from '@accelint/core';
```
kalisjoshua, line 2: Just a question, why do we need a library function for a `noop`? I saw it in there and I was very confused.

Additionally, why would we take ownership of implementations in other libraries we use?

packages/design-system/src/hooks/use-tree/use-tree.ts
```diff
@@ -1,4 +1,4 @@
-import { isEqual } from 'lodash';
+import isEqual from 'lodash/isEqual';
```
kalisjoshua, line 1: Is this better for tree-shaking? Just curious about the change.

kalisjoshua (approved):

packages/design-system/src/hooks/use-tree/use-tree.ts
brandonjpierce: Somewhat of a temp fix as we investigate this error more closely which occurred in one of our consuming apps of the design system package

image

brandonjpierce: But without extra tooling, yes, it is technically better to do import like this for lodash for treeshaking purposes

brandonjpierce (approved):

packages/design-system/src/components/button/button.tsx
kalisjoshua: I'm not suggesting a change at all, I am genuinely curious. From my perspective what I see is that with a `noop` as part of a library:

1. We have it to maintain - not that there is much to do for this to be clear
2. We have to import it to use it which is more laborious than just writing it

```
import { noop } from '@accelint/core'; // 38 characters
const noop = () => {}; // 22 characters
```

3. Additionally it is added footprint for the library:
    - Download + parse etc.
    - Dependency graph management

I just don't see the upside. I don't see much harm to be clear just raising it as a topic to discuss because I am new still.

brandonjpierce: https://github.com/gohypergiant/standard-toolkit/discussions/83 for convo above

belsrc: I'll "pick" on point 2. Normal D.R.Y. things.  I don't want to remake it that many times (image). And the laborious is situational. Chances are, other items are also being used so then the count is just `5` (`6` if you count return)
Screenshot 2024-11-05 at 9 51 50 AM.

PR #87, opened 2024-11-05
Switch to peer deps, handle Lodash imports with plugin

ArrayKnight:


brandonjpierce:
@ArrayKnight what are your thoughts on using exclusively the `@react-aria/*` packages instead of the hoisted `react-aria` package? Would help out with things like tree shaking and for downstream tooling like Vitest, etc.

ArrayKnight:
> @ArrayKnight what are your thoughts on using exclusively the `@react-aria/*` packages instead of the hoisted `react-aria` package? Would help out with things like tree shaking and for downstream tooling like Vitest, etc.

I've switched it over to this. It expands the peer dep list and is a little bit more cumbersome to use. I have to reference https://github.com/adobe/react-spectrum/blob/main/packages/react-aria/src/index.ts to figure out where to import things from

brandonjpierce (approved):

PR #88, opened 2024-11-06
feat(geo/src/coordinates): add coordinate parsing and validation

### LINKED ISSUE #70: Finish coordinate suite
1. Finish the coordinate parser regex and edge case detection

This is definitely the most complex part of the suite. The general ethos around the parsing is that we try and make our capture groups as deterministic as humanly possible. Less tricks in the regex the better for simplification sake. E.g. for a coordinate passed to us in DD form we will always capture the following groups, regardless if there is a value there or not:

```
[dir][sign][degrees][decimals][dir][sign]
``` 

With this, we can capture either lon or lat independently, or combined by simply concating the regex's. Additionally, we now have more deterministic inputs for our value normalization described further below.

The current Regex partials can be found here:

```
packages/geo/src/coordinates/regex.ts
packages/geo/src/coordinates/match.ts
```

We currently are testing against a number of variations for each coordinate format as seen here:

```
packages/geo/src/coordinates/__fixtures__/index.ts
```

This covers the bulk of the edge cases and or use cases that you would see in the wild. We will want to slightly adjust these fixtures potentially just so that the snapshot output is a tad more consistent and or our `expect()` calls can be simplified a bit or in a more generic way.

2. Finish the coordinate normalization

Given that we now have the input, either from a machine or a user, deterministic split into N number of capture groups, we can normalize this input into a common form. Notably, regardless of the input format, we want the normalized format to be a simple float value that can then be used as the input for external formatting outwards. E.g. `N 87.2123` would become `87.2123` , `87.2123 S` would become `-87.2123` etc.

We have some initial work started for this for specifically the `DD` format capture groups here:

```
packages/geo/src/coordinates/normalize.ts
```

3. Create coordinate formatting

Given our normalized float value for both lon and lat we can take that and format it into the common standards of DD, DMS, DDM. We _might_ be able to actually reuse the regex segments to assist with this but this area is largely tbd. Additionally, we will potentially need to support MGRS, CGRS, and other standard military coordinate formats.

4. Create coordinate component

kalisjoshua:
closes #70


packages/geo/src/patterning.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+const nomatch = (t: string) => new RegExp(`___UNMATCHED_${t}_TOKEN___`);
+
+export const Patterning = {
```
belsrc, line 16: This object wrapping isn't really necessary in a module based system. The module itself is your object and your encapsulation. And forcing the grouping can have downstream implications.

More of an "open a discussion" addition, I'm very much one for *not* adding dep's willy-nilly but I believe this is just for constructing regex's (from a brief skim). Would it be better long term to just use something like [ts-regex-builder](https://www.npmjs.com/package/ts-regex-builder), [super-expressive](https://github.com/francisrstokes/super-expressive) or similar?

belsrc (commented):
1/?

packages/geo/src/coordinates/latlon/internal/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  Separator = '/',
+}
+
+export const BEARINGS = {
+  // [positive, negative] so that "isNegative" can be used to index into the array
+  LAT: [Compass.North, Compass.South],
+  LON: [Compass.East, Compass.West],
+} as const;
+
+export const DEFAULT_FORMAT: Format = 'LONLAT';
```
brandonjpierce, line 47: Regarding a default ordering, we actually should do `LATLON` as that is what operators and such are use to (even though it is `LONLAT` in some specs like geojson).

packages/geo/src/coordinates/latlon/internal/parse.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    getViolations,
+
+    (tokens: Tokens) => {
+      if (!(violations.length || tokens.includes(Symbols.Separator))) {
+        addDividers(tokens);
+      }
+    },
+
+    repositionBearings,
```
brandonjpierce, line 49: If possible lets try and avoid mutation on args and or in general and utilize side effect free functions. This makes it easier to test functions in isolation imo and opens the door for really nice composition e.g.

```js
const tokens = lexer(s);

// could use compose() as well and reverse fn ordering
const output = pipe(
  getConflicts,
  addMissingBearings,
  repositionBearings,
)(tokens);
```

packages/geo/src/coordinates/latlon/internal/index.ts
belsrc: Personally, I would say "all" as it correlates to `x, y` /pedantic. My argument to that is, I would rather be correct as a whole than to change that correctness for in internal use case.

packages/geo/src/coordinates/latlon/internal/parse.test.ts
```diff
@@ -0,0 +1,66 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { expect, it, describe } from 'vitest';
+
+import { parse } from './parse';
+
+describe('parser', () => {
+  it.each`
```
brandonjpierce, line 19: Minor nit with the `it.each` approach: if we wanted to use these pairs in other tests, for mocks, for fixtures, etc we would have to copy paste this block around. Putting the pairs (value, expectation) into a fixture means we can use it all over the place if needed.

packages/geo/src/coordinates/latlon/internal/lexer.ts
```diff
@@ -0,0 +1,61 @@
...
+        Patterning.optional(SYMBOL_PATTERNS.dms),
+      ),
+    ),
+    'gi',
+  ),
+};
+
+export function lexer(string: string) {
+  const tokens =
+    string
```
brandonjpierce, line 50: Minor nit: since `String` is a reserved word and `string` is a ts type, can we swap to a var name of `value` or something generic?

packages/geo/src/coordinates/latlon/internal/lexer.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { Patterning } from '@/patterning';
+
+import { SYMBOL_PATTERNS, Symbols } from '.';
+
+export type Tokens = ReturnType<typeof lexer>;
+
+const patterns = {
+  positive: /\+/g,
+  separators: /[,/]/g,
```
brandonjpierce, line 22: We may want to leave a note here or future TODO to support a space as a separator (will sanity check if that is a common scenario in the field first)

packages/geo/src/coordinates/latlon/internal/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export const LIMITS: Record<Format, [Limits, Limits]> = {
+  LATLON: [Limits.Lat, Limits.Lon],
+  LONLAT: [Limits.Lon, Limits.Lat],
+};
+
+export const SYMBOL_PATTERNS = {
+  LAT: new RegExp(`[${BEARINGS.LAT.join('')}]`),
```
brandonjpierce, line 55: A fun thing we did at the very start of this endeavor for the sake of other devs was added in https://regex-vis.com/ links as comments above the regex so folks could see an explanation of it. Might be a good idea to pop those in as well during the polish stage. E.g. https://regex-vis.com/?r=%2F%5Cs*%2F

brandonjpierce: Another site with a slightly different visual style: https://www.nodexr.net/?parse=(%5B-%2B%5D)%5Cs*

belsrc: Oh! I like those. ( `{@link https://www.nodexr.net/?parse=(%5B-%2B%5D)%5Cs }`)

packages/geo/src/coordinates/coordinate.ts
```diff
@@ -0,0 +1,281 @@
...
+type Value = {
+  latitude: number;
+  longitude: number;
+};
+
+enum Systems {
+  DecimalDegrees = 'Decimal Degrees',
+  DegreesDecimalMinutes = 'Degrees Decimal Minutes',
+  DegreesMinutesSeconds = 'Degrees Minutes Seconds',
+}
```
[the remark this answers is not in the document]
kalisjoshua: I've removed the use of `enum` in favor of objects. I'll finish some other changes based on feedback before pushing changes.

packages/geo/src/patterning.ts
kalisjoshua: I converted these over to exporting individual function instead. I'll finish some other changes based on feedback before pushing changes.

packages/geo/src/coordinates/latlon/internal/parse.ts
kalisjoshua: removed.

packages/geo/src/coordinates/latlon/decimal-degrees/parser.ts
```diff
@@ -0,0 +1,112 @@
...
+
+function identifyErrors(format: Format) {
+  return (arg: DecimalDegrees | undefined, i: number) => {
+    if (!arg) {
+      return [[], ['Invalid coordinate value.']] as ParseResults;
+    }
+
+    let { bear, deg } = arg;
+
+    deg ||= '0';
```
brandonjpierce, line 68: Sanity checking the `||=` here vs a `deg ?? '0'`

packages/geo/src/coordinates/latlon/internal/parse.test.ts
kalisjoshua: Yeah, I can see that. However, it is very nice to be able to see the input an expected output at a glance. Is this something we want to change?

brandonjpierce: Looks like we can maybe leverage `it.for` https://vitest.dev/api/#test-for

```
const tests = [
  ['0 0', '0 / 0', [],
  ['.01 .02', '0.01 / 0.02', []
]

it.for(tests)(
  '%i | %i | %i', 
  [a, b, errors] => {
    // expect
  }
);

```

That way we can pass around `tests` still if needed for other things

kalisjoshua: Either would work. Do you want me to make that change now or can it wait till we need to reuse test cases? The only thing I hang onto is that the template string gets auto-formatted for column layouts #lazy-dev.

kalisjoshua:
> 1/?

What does this mean? What is it asking?

kalisjoshua:
Is this good to merge? Are there things I should change before merge?

brandonjpierce:
This is likely good to go for our first implementation pass to test things in the wild. We may end up slightly adjusting the folder/file structure but that can happen later once we have more things in `@accelint/geo`. Thanks for all of the hard work on this one!

brandonjpierce (dismissed):

brandonjpierce:
@kalisjoshua actually one final TODO: add in the changeset

belsrc:
Just run the `changeset` script in the root package file. It will run an interactive CLI that walks you through it. Will pick which packages were changed, which (if any) were `major`. Will pick which packages were changed, which (if any) were `minor`. Will pick which packages were changed, which (if any) were `patch`. And will enter a description of change. Once done, it will create a markdown file of what you entered.

packages/geo/package.json
```diff
@@ -1,12 +1,10 @@
 {
   "name": "@accelint/geo",
-  "version": "0.1.3",
+  "version": "1.0.0",
```
brandonjpierce, line 3: We may want to change this back to `0.0.0` as the `major` changeset would actually bump this to `2.0.0`. Additionally, for a initial release, we have been going with a `minor` version to sort of denote a "beta" in a way. e.g. `0.1.0`

brandonjpierce: Could certainly do that as well! Given we are starting from "scratch" a little bit with the new implementation, I am also okay with a `0.1.0`

brandonjpierce: Actually, just remembered we were publishing this already to NPM so we will need to keep the current version of `0.1.3`

brandonjpierce (approved):

PR #89, opened 2024-11-06
fix: ls-lint

brandonjpierce:
- Update config to remove `pointcase` which was problematic
- Update config to use glob pattern for `pointcase` scenarios
- Update some of the ignored folders
- Support `screaming_snake_case` for markdown files to handle `README.md`, etc

Validated configuration via `pnpm ls-lint --debug`


package.json
```diff
@@ -33,7 +33,7 @@
     "@biomejs/biome": "^1.9.3",
     "@changesets/changelog-git": "^0.2.0",
     "@changesets/cli": "^2.27.9",
-    "@ls-lint/ls-lint": "^2.2.3",
+    "@ls-lint/ls-lint": "2.3.0-beta.1",
```
brandonjpierce, line 36: This was needed to support the wildcard extensions (pointcase fix)

kalisjoshua (approved):

Issue #103, opened 2024-11-17
Audit code docblocks for documentation generation


belsrc:
We need to audit the code that is currently in the code base so that when the documentation is generated it doesn't have "holes" in it.

The docblock should have a minimum of a description and at least one `@example`. `@param` and `@return` are not needed as the types themselves should cover it. Unless they are particularly complex.

PR #106, opened 2024-11-18
feat(lint): check for docblock in public API exports

### LINKED ISSUE #113: Create script to check correctness of docblocks
### Provide a general summary of the feature

Create a script that we can then tie into Turbo and CI pipeline that would check the code for an individual package and inform the developer whether or not all publicly exposed items have a well structured docblock.

### 🤔 Expected Behavior?

The script should run on the all of the code files in a package. It could also respect the `// @export-*` comments that will come with the indexer CLI. It should make sure that the docblock has a minimum of a description and at least one `@example`. `@param` and `@return` are not needed as the types themselves should cover it. Unless they are particularly complex.

### 😢 Current Behavior

_No response_

### 🤩 Possible Solution

_No response_

### 📜 Context

More automation and better resulting documentation.

### 💻 Examples

_No response_

kalisjoshua:
Closes #113

## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No

## 💬 Other information

issue #113


package.json
```diff
@@ -25,6 +25,7 @@
     "indexer": "turbo index",
     "license": "pnpm zx scripts/license.mjs",
     "lint": "turbo lint",
+    "postlint": "pnpm zx scripts/docblock.mjs",
```
belsrc, line 28: Can probably just make this part of the lint command so that turbo can cache it.

scripts/docblock.mjs
```diff
@@ -0,0 +1,136 @@
...
+function exportsReducer(acc, node) {
+  if (node.type === 'ExportDeclaration') {
+    if (node.declaration.identifier) {
+      acc.push(node.declaration.identifier.value);
+    } else if (node.declaration.declarations) {
+      for (const inner of node.declaration.declarations) {
+        acc.push(inner.id.value);
+      }
+    }
+  }
```
belsrc, line 33: Type exports are also `declaration.id.value`

scripts/docblock.mjs
```diff
@@ -0,0 +1,136 @@
...
+      break;
+    default:
+      result.push(exports);
+      break;
+  }
+
+  return result;
+}
+
+const pragmaParser = (() => {
```
belsrc, line 64: Readability nit: Can we get this fn above `getFileDetails` since that is the calling order

kalisjoshua: I try and rely on objective ordering this is in alpha order rather than anything subjective. I can move it around if necessary though.

scripts/docblock.mjs
kalisjoshua: For the context - docblock - does it matter? Do we want to expect docblock for type alias'?

package.json
kalisjoshua: Does turbo not execute `post*` script automatically?

scripts/docblock.mjs
belsrc: Yea. Still a part of the public facing API. And, depending on how the type is constructed, intellisense won't help. Unless you "goto definition" several times.

package.json
belsrc: By natural of it running package commands, it might. But I don't think that it will cache the output which is needed to allow it to "skip" processes that don't need to run again.

kalisjoshua: How do we test it to see?

scripts/docblock.mjs
belsrc: Every human reads top-to-bottom so it makes it less work (scroll up and down, "was that an import", scroll up, "nope, where is it defined then", scroll down, back and forth) to understand what is going on in examples like this. Where it is basically procedural. That being said, content like constant files, types files or function files with no internal dependencies I don't have an issue with them being alpha order.

kalisjoshua: I broke them out into separate modules to make it cleaner overall; and potentially we could reuse pieces.

scripts/docblock.mjs
kalisjoshua: I think I got this fixed. Let me knwo.

package.json
kalisjoshua: I moved it into `lint`.

belsrc: Yea, doing `&& command` still won't cache it in turbo. Need to do something like [Adding tasks](https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#defining-tasks) and then in the actual `lint` script do something like [running multiple](https://turbo.build/repo/docs/crafting-your-repository/running-tasks#running-multiple-tasks)

kalisjoshua: I added the task to the turbo config - https://github.com/gohypergiant/standard-toolkit/pull/106/files#diff-f8de965273949793edc0fbfe249bb458c0becde39b2e141db087bcbf5d4ad5e3R10-R13.

belsrc:
Additionally, @brandonjpierce, I'm assuming we want this to be per package like normal? For caching reasons.

belsrc:
Since it is relevant, this doesn't check for "correctness" does it? It's purely "do I have at least a `/** */` above the code?"

kalisjoshua:
> Since it is relevant, this doesn't check for "correctness" does it? It's purely "do I have at least a `/** */` above the code?"

I thought that was what we talked about? Checking for "correctness" is a deeper rabbit hole; one I am happy to go down if that is what we want.

kalisjoshua:
> Ok, then let me spin up a new ticket as this doesn't directly address the one linked in the description. As for the second part of your comment, personally I think "probably". I don't know the usefulness of simply checking if there is any comment since I'm assuming `/** */` will "pass". Which wouldn't do anything to help us with docs.

The ticket doesn't detail what it is looking for. What defines "correctness"?

belsrc:
Added #113. Updated this description for link. Clarified in original #103.

To add to that, most of the issues that me and Brandon created were done in like 30min and were aimed towards us. So if there is additional clarity needed for you, please just pop a comment in them.

kalisjoshua:
If we are just looking for the docblock to have those elements - description and example - then that is easy enough. But executing the examples to check that level of correctness would be more involved.

PR #108, opened 2024-11-20
fix(tests): use vitest .each convention for tests

### LINKED ISSUE #11: Vitest conventions
A common pattern we have is doing something like:

```
const pairs = [
  ['test title', 123],
  ['test title', 123],
  ['test title', 123],
  ['test title', 123],
]

it('the test', () => {
  for (const pair of pairs) {
    expect(pair[1]).toEqual(123);
  } 
});
```

However, vitest does have `it.each` which works on pairs like datasets. Should we switch to that method instead?

kalisjoshua:
Closes #11 

## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/converters/src/boolean-to-number/index.test.ts
+++ b/packages/converters/src/boolean-to-number/index.test.ts
@@ -13,15 +13,12 @@
 import { describe, expect, it } from 'vitest';
 import { booleanToNumber } from './';
 
-const testPairs: [boolean, number][] = [
-  [true, 1],
-  [false, 0],
-];
-
 describe('bool to number', () => {
-  for (const [input, result] of testPairs) {
-    it(`should convert ${input} to ${result}`, () => {
-      expect(booleanToNumber(input)).toEqual(result);
-    });
-  }
+  it('should convert true to 1', () => {
+    expect(booleanToNumber(true)).toEqual(1);
+  });
+
+  it('should convert false to 0', () => {
+    expect(booleanToNumber(false)).toEqual(0);
+  });
 });
--- a/packages/predicates/src/is-number/index.test.ts
+++ b/packages/predicates/src/is-number/index.test.ts
@@ -12,152 +12,69 @@
 
 import { expect, it, describe } from 'vitest';
 import { isFiniteNumber, isFiniteNumeric, isNumber, isNumeric } from './';
-import {
-  floatingPointPairs,
-  floatingPointStringPairs,
-  nonFinitePairs,
-  nonFiniteStringPairs,
-  nonNumericPairs,
-  numberLikeStringPairs,
-  numberLiteralPairs,
-} from './__fixtures__';
+
+// biome-ignore lint/style/useNumberNamespace: want to diff against Number.NEGATIVE_INFINITY
+const INFINITY = Infinity;
+const { POSITIVE_INFINITY, NEGATIVE_INFINITY } = Number;
 
 describe('number validators', () => {
-  for (const pairs of numberLiteralPairs) {
-    it(`isFiniteNumber: ${pairs[0]}`, () => {
-      expect(isFiniteNumber(pairs[1])).toBeTruthy();
-    });
+  it.each`
+    label                                      | value                | isFiniteNumber | isFiniteNumeric | isNumber | isNumeric
+    ${'integer literal: negative'}             | ${-1}                | ${true}        | ${true}         | ${true}  | ${true}
+    ${'integer literal: zero'}                 | ${0}                 | ${true}        | ${true}         | ${true}  | ${true}
+    ${'integer literal: positive'}             | ${1}                 | ${true}        | ${true}         | ${true}  | ${true}
+    ${'integer literal: octal'}                | ${0o0144}            | ${true}        | ${true}         | ${true}  | ${true}
+    ${'integer literal: hexadecimal'}          | ${0xfff}             | ${true}        | ${true}         | ${true}  | ${true}
+    ${'floating point: nagative'}              | ${-1.1234}           | ${true}        | ${true}         | ${true}  | ${true}
+    ${'floating point: positive'}              | ${1.1234}            | ${true}        | ${true}         | ${true}  | ${true}
+    ${'floating point: exponential notation'}  | ${8e5}               | ${true}        | ${true}         | ${true}  | ${true}
+    ${'number-like string: negative'}          | ${'-1'}              | ${false}       | ${true}         | ${false} | ${true}
+    ${'number-like string: zero'}              | ${'0'}               | ${false}       | ${true}         | ${false} | ${true}
+    ${'number-like string: positive'}          | ${'1'}               | ${false}       | ${true}         | ${false} | ${true}
+    ${'number-like string: octal'}             | ${'040'}             | ${false}       | ${true}         | ${false} | ${true}
+    ${'number-like string: hexadecimal'}       | ${'0xFF'}            | ${false}       | ${true}         | ${false} | ${true}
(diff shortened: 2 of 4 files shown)


packages/converters/src/boolean-to-number/index.test.ts
```diff
@@ -13,15 +13,12 @@
...
-    it(`should convert ${input} to ${result}`, () => {
-      expect(booleanToNumber(input)).toEqual(result);
-    });
-  }
+  it.each`
+    input    | result
+    ${true}  | ${1}
+    ${false} | ${0}
+  `('should convert $input to $result', ({ input, result }) => {
+    expect(booleanToNumber(input)).toEqual(result);
```
belsrc, line 22: Prefer not to use string template: https://github.com/gohypergiant/standard-toolkit/pull/88#discussion_r1834744892

packages/converters/src/to-boolean/index.test.ts
```diff
@@ -10,34 +10,21 @@
...
-  for (const item of falsey) {
-    it(`should return false for ${item}`, () => {
-      expect(toBoolean(item)).not.toBeTruthy();
+describe('toBoolean', () => {
+  describe.each([
+    // positive cases
+    ...createFixture(true, [truthy, toBoolean]),
+    // negative cases
+    ...createFixture(false, [falsey, toBoolean]),
+  ])('%s', (_name, expected, fn, values) => {
```
belsrc, line 25: Yea not just export this completed from the fixture? Test should minimize as much logic as possible.

belsrc:
As it looks like you doing like 5 things in this one PR. Please try to keep a PR to a single issue.

packages/converters/src/to-boolean/index.test.ts
kalisjoshua: It's custom per test; `createFixture` is used in more than one place.

packages/converters/src/boolean-to-number/index.test.ts
belsrc: That's subjective. Array, or object, allows much more flexibility and less typing (in the case of array).

kalisjoshua:
> As it looks like you doing like 5 things in this one PR. Please try to keep a PR to a single issue.

I did let things get away from me on this one. I can break things up they are pretty small. Should I?

packages/converters/src/to-boolean/index.test.ts
kalisjoshua: Should I move it down and reuse across modules at the lower level?

belsrc:
> I did let things get away from me on this one. I can break things up they are pretty small. Should I?

If you don't mind I would appreciate it. Just makes it easier to discuss each thing individually without context switching and makes it a lot easier to track work.

packages/converters/src/to-boolean/index.test.ts
belsrc: Personally, I subscribe heavily to [the kind of "go-to" best practices doc](https://github.com/goldbergyoni/javascript-testing-best-practices?tab=readme-ov-file#section-0%EF%B8%8F%E2%83%A3-the-golden-rule) so I don't really consider stuff like that needed. Which dovetails into the other discussion of making it dead simple, colocated and easier to maintain.

kalisjoshua: Duplicating code is a bad practice too, and worse than not having everything colocated. #IMHO

kalisjoshua:
I believe this is only making changes to tests at this point. I moved the common fixtures to the `constants/` module because that felt more natural and consistent; to have the consumers both referencing it from there rather than one referencing it in another module.

packages/constants/src/__fixtures__/boolean.ts
```diff
@@ -0,0 +1,45 @@
...
+  'TRUE',
+];
+export const falsey: Values = [
+  0,
+  '0',
+  'off',
+  'false',
+  'no',
+  false,
+  [],
```
ArrayKnight, line 37: I don't think that `[]` or `{}` are something that we can safely consider falsey.

It feels like you're trying to suggest that only empty versions of these objects would considered falsey(?)

But I'm pretty sure that if you put a populated version into the truthy values, it would fail.

kalisjoshua: I do not think that they should be considered as either "false" or "true" personally. However, that is what was in the source before I got to it so I was keeping them for consistency only. https://github.com/gohypergiant/standard-toolkit/pull/108/files#diff-49742f138b932b2304849053f633a630876e6a9d469f2fb6cc7d3d9351378024L24-L25

ArrayKnight: I tend to agree that they shouldn't be in either group. It does feel like something that should be fixed

kalisjoshua: I'd be happy to refactor these out if that is the direction we want to go... but in another PR.

ArrayKnight: Yea: https://github.com/gohypergiant/standard-toolkit/pull/108/files#diff-9e7c7e0fa586596636eafa34b75c343bcc48590b97395798c7d78dca3aea8ea1R15 this is problematic since any object will stringify to that value

kalisjoshua: I did note this slightly in the code; https://github.com/gohypergiant/standard-toolkit/pull/108/files#diff-9e7c7e0fa586596636eafa34b75c343bcc48590b97395798c7d78dca3aea8ea1R13-R14.

packages/converters/src/to-boolean/index.test.ts
```diff
@@ -10,34 +10,25 @@
...
-  {},
-  'OFF',
-  'NO',
-  'FALSE',
-];
+import {
+  createFixture,
+  falsey,
+  truthy,
+} from '../../../constants/src/__fixtures__/boolean';
```
brandonjpierce, line 19: I am on the fence on this one for a few reasons:

1. We are bridging packages somewhat _incorrectly_ here, if you are referencing a package import like `@accelint/constants` this means you are technically referencing "compiled" artifacts not raw source code. 
2. The line in the sand on what does or does not constitute a "shared" or "constant" fixture will become muddy over time (especially as newer folks onboard). These checks are so incredibly simple I don't mind some minor copy paste.
3. If we absolutely must have a centralized spot for fixtures, which is something we do on one of our projects, we should just make it a formal package e.g. `@accelint/fixtures`. This does have some of the issues raised in point 2 but at the very least allows us to "compile" things like a normal package and import them as such as well.

Some fun reading regarding my thoughts around reuse, copy paste, and abstractions:

- http://programmingisterrible.com/post/139222674273/write-code-that-is-easy-to-delete-not-easy-to
- https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

kalisjoshua: I think this is largely moot after the refactor of `toBoolean` and `is-noyes` (predicates) lands.

kalisjoshua: Good reads by the way.

kalisjoshua: Removed after the `toBoolean` refactor.

packages/converters/src/to-boolean/index.test.ts
kalisjoshua: This no longer exists after the refactor of `toBoolean`.

brandonjpierce (approved):

PR #110, opened 2024-11-21
feat: replace regex with value lookup

### LINKED ISSUE #77: PERF: Change yesno predicate away from regex
The current implementation of the various funcs in `predicates/is-noyes` could be altered to increase performance.

```js
const toBool = (val) =>
  val === 1 ||
  val === '1' ||
  val === 'on' ||
  val === 'true' ||
  val === 'yes' ||
  val === true;
  
// --------------

function normalize(val) {
  return `${val}`.trim();
}
const trueRegex = /^(?:y|yes|true|1|on)$/i;
function isTrue(val) {
  const normalized = normalize(val);
  return trueRegex.test(normalized);
}

// --------------

const truthy = [1, '1', 'on', 'true', 'yes', true, 'ON', 'YES', 'TRUE'];
function inBool(val) {
  return truthy.includes(val);
}

// --------------

const test = [1, 0, '1', '0', 'on', 'off', 'true', 'false', 'yes', true, false, 'ON', 'YES', 'TRUE'];

// --------------

// 4.1M ops/s ± 0.88%
// 3.03 % slower
test.forEach(x => toBool(x));

// --------------

// 2.4M ops/s ± 0.78%
// 42.59 % slower
test.forEach(x => isTrue(x));

// --------------

// 4.2M ops/s ± 1.12%
// Fastest
test.forEach(x => inBool(x));
```

Additionally the negations can just use `not` instead and the other variations can just be aliases.

```ts
// export function isFalse(val: unknown) {
//   const normalized = normalize(val);
//   return falseRegex.test(normalized);
// }

export const isFalse = notFn(isTrue);

// export function isNo(val: unknown) {
//   return isFalse(val);
// }

export const isNo = isFalse;

// export function isYes(val: unknown) {
//   return isTrue(val);
// }

export const isYes = isTrue;
```

kalisjoshua:
Closes #77 

## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/predicates/src/is-noyes/index.bench.ts
+++ b/packages/predicates/src/is-noyes/index.bench.ts
@@ -1,0 +1,326 @@
+/*
+ * Copyright 2024 Hypergiant Galactic Systems Inc. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { bench, describe } from 'vitest';
+
+// biome-ignore lint/suspicious/noExplicitAny: <explanation>
+type Callable = (...a: any[]) => any;
+
+const TEST_VALUES = [
+  1,
+  0,
+  '1',
+  '0',
+  'on',
+  'off',
+  'true',
+  'false',
+  'yes',
+  true,
+  false,
+  'ON',
+  'YES',
+  'TRUE',
+] as const;
+
+const runTestValues = (fn: Callable) => TEST_VALUES.forEach(fn);
+
+// NOTE: doing a bench() inside of the for/of loop causes heap allocation errors
+describe('variations on isFalse|isNo|isOff|isOn|isTrue|isYes implementation', () => {
+  /**/ // <--- remove space to disable block comment
+  bench('original (regex) with setup', () => {
+    const trueRegex = /^(?:y|yes|true|1|on)$/i;
+    const falseRegex = /^(?:n|no|false|0|off)$/i;
+
+    function isFalse(val: unknown) {
+      const normalized = normalize(val);
+      return falseRegex.test(normalized);
+    }
+
+    function isTrue(val: unknown) {
+      const normalized = normalize(val);
+      return trueRegex.test(normalized);
+    }
+
+    function normalize(val: unknown) {
+      return `${val}`.trim();
+    }
+
+    runTestValues(isTrue);
+    runTestValues(isFalse);
+  });
+  //*/
+
+  /** / // <--- remove space to disable block comment
+  (() => {
+    // isolated scope
+    const trueRegex = /^(?:y|yes|true|1|on)$/i;
+    const falseRegex = /^(?:n|no|false|0|off)$/i;
+
+    function isFalse(val: unknown) {
+      const normalized = normalize(val);
+      return falseRegex.test(normalized);
+    }
+
+    function isTrue(val: unknown) {
+      const normalized = normalize(val);
+      return trueRegex.test(normalized);
+    }
+
+    function normalize(val: unknown) {
+      return `${val}`.trim();
+    }
+
+    bench('original (regex) without setup', () => {
+      runTestValues(isTrue);
+      runTestValues(isFalse);
+    });
+  })();
+  //*/
+
+  /**/
+  bench('[altered] original (regex) with setup', () => {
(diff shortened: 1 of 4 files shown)


belsrc:
That actually turns out to be slower than the original regex version.

```js
// -----------------------------------
// Setup

const test = [1, 0, '1', '0', 'on', 'off', 'true', 'false', 'yes', true, false, 'ON', 'YES', 'TRUE'];

function normalize(val) {
  return `${val}`.trim();
}

function normalize2(val) {
  switch ({}.toString.call(val)) {
    case '[object Array]':
      return `${val.length}`;
    case '[object Object]':
      return `${Reflect.ownKeys(val).length}`;
    default:
      return `${val}`.trim().toLowerCase();
  }
}

// -----------------------------------
// Test 1
// 1.5M ops/s ± 3.22%
// 54.99 % slower

const VAL_TRUE = [1, true, 'y', 'yes', 'on'].map(normalize2);

const isTrue = (val) => VAL_TRUE.includes(normalize2(val));

test.forEach(x => isTrue(x));

// -----------------------------------
// Test 2
// 2.3M ops/s ± 1.34%
// 33 % slower

const trueRegex = /^(?:y|yes|true|1|on)$/i;

function isTrue(val) {
  const normalized = normalize(val);
  return trueRegex.test(normalized);
}

test.forEach(x => isTrue(x));

// -----------------------------------
// Test 3
// 3.4M ops/s ± 1.14%
// Fastest

const truthy = [1, '1', 'on', 'true', 'yes', true, 'ON', 'YES', 'TRUE'];

function inTrue(val) {
  return truthy.includes(val);
}

test.forEach(x => inTrue(x));
```

Even if we take best case scenario where we ignore the time to initialize.

```js
// -----------------------------------
// Setup

const test = [1, 0, '1', '0', 'on', 'off', 'true', 'false', 'yes', true, false, 'ON', 'YES', 'TRUE'];

function normalize(val) {
  return `${val}`.trim();
}

function normalize2(val) {
  switch ({}.toString.call(val)) {
    case '[object Array]':
      return `${val.length}`;
    case '[object Object]':
      return `${Reflect.ownKeys(val).length}`;
    default:
      return `${val}`.trim().toLowerCase();
  }
}

const VAL_TRUE = [1, true, 'y', 'yes', 'on'].map(normalize2);
const isTrue1 = (val) => VAL_TRUE.includes(normalize2(val));

const trueRegex = /^(?:y|yes|true|1|on)$/i;
function isTrue2(val) {
  const normalized = normalize(val);
  return trueRegex.test(normalized);
}

const truthy = [1, '1', 'on', 'true', 'yes', true, 'ON', 'YES', 'TRUE'];
function inTrue(val) {
  return truthy.includes(val);
}

// -----------------------------------
// Test 1
// 2.6M ops/s ± 0.4%
// 54.13 % slower

test.forEach(x => isTrue1(x));

// -----------------------------------
// Test 2
// 3.2M ops/s ± 0.36%
// 43.54 % slower

test.forEach(x => isTrue2(x));

// -----------------------------------
// Test 3
// 5.6M ops/s ± 0.38%
// Fastest

test.forEach(x => inTrue(x));
```

kalisjoshua:
So `normalize` is definitely slowing things down a little; that's not a surprise. The array lookup `includes` being slower that an regex isn't all that surprising either. I've got some ideas to try out and compare; I will attempt a bench file.

kalisjoshua:
I see, it's the normalize I wrote that is slower.

kalisjoshua:
I created a few options in a bench file for no|yes; I think the array lookup is now the "best" solution. There is a faster solution but I don't think it is the best from a maintainability perspective; if we need to squeeze the absolute most performance out though, it is faster.

packages/predicates/src/is-noyes/index.bench.ts
```diff
@@ -0,0 +1,326 @@
...
+    });
+  })();
+  //*/
+
+  /**/
+  bench('[altered] original (regex) with setup', () => {
+    const trueRegex = /^(?:y|yes|true|1|on)$/i;
+    const falseRegex = /^(?:n|no|false|0|off)$/i;
+
+    const test = (regex: RegExp, val: unknown) => regex.test(`${val}`.trim());
```
brandonjpierce, line 95: We may want to hoist this function and potentially the regex outside of `bench` to remove any overhead from function allocation etc (i'm not sure if values created in a `bench` callback are created each run of of the bench)

belsrc: More than likely are, third "best case" scenario that I linked above. (Though it is from a different platform)

brandonjpierce:
Outside of the potential `bench` adjustment, LGTM 👍

brandonjpierce (approved):

kalisjoshua:
For posterity's sake here are the numbers; on my machine.

image

packages/predicates/src/is-noyes/index.bench.ts
belsrc: Yea, I find "setup" to still be partially relevant since it does change start-up times. I usually run two benchmarks side-by-side to make sure that it isn't egregious. But I also don't know how we want to do it in this repo.

brandonjpierce:
@kalisjoshua `patch` level changeset if you don't mind

brandonjpierce (approved):

PR #111, opened 2024-11-22
fix: use semver for workspace packages

### LINKED ISSUE #34: Swap to semver ranges for workspace packages
Move away from `workspace:*` for local workspace package dependencies. Reasoning is that we want a bit tighter control over what dependency is used for a package regardless where it is coming from. You could, for instance, creating a breaking semver change in a workspace package that is a dependency of another workspace package and inadvertently break that package or force yourself into making that one a breaking semver change as well.

kalisjoshua:
Closes #34

## ✅ Pull Request Checklist:

- [ ] Yes
- [ ] No


kalisjoshua:
I'm not sure why the design-system package isn't working the same as the others.

ArrayKnight:
> I'm not sure why the design-system package isn't working the same as the others.

Can you expand on that? What does it do differently? Can't tell from the code alone

kalisjoshua:
When the version was specified `0.4.0` instead of `workspace:*` it fails to build in CI - https://github.com/gohypergiant/standard-toolkit/actions/runs/11978494290/job/33401372516 - but I could not reproduce locally.

belsrc:
Might be relevant ... which I'm assuming the later `workspace:` is preferred. Though depending on how changesets interplay with it, may need to go the other route.

> If [link-workspace-packages](https://pnpm.io/npmrc#link-workspace-packages) is set to true, pnpm will link packages from the workspace if the available packages match the declared ranges. For instance, foo@1.0.0 is linked into bar if bar has "foo": "^1.0.0" in its dependencies and foo@1.0.0 is in the workspace. However, if bar has "foo": "2.0.0" in dependencies and foo@2.0.0 is not in the workspace, foo@2.0.0 will be installed from the registry. This behavior introduces some uncertainty.
>
> Luckily, pnpm supports the workspace: protocol. When this protocol is used, pnpm will refuse to resolve to anything other than a local workspace package. So, if you set "foo": "workspace:2.0.0", this time installation will fail because "foo@2.0.0" isn't present in the workspace.
>
> This protocol is especially useful when the [link-workspace-packages](https://pnpm.io/npmrc#link-workspace-packages) option is set to false. In that case, pnpm will only link packages from the workspace if the workspace: protocol is used.

brandonjpierce:
Per Bryan's note, to ensure that local packages are only used we still need to specify the workspace protocol but use specific versions instead of the `*` e.g.

```
"@accelint/vitest-config": "workspace:0.1.3"
```

As specifying as

```
"@accelint/vitest-config": "0.1.3"
```

Is grabbing the remote published package off of npm.

brandonjpierce:
Since this doesn't impact the actual published code, no need for a changeset imo.

PR #116, opened 2024-12-03
refactor: make anys explicit

### LINKED ISSUE #115: refactor: Clean up linter warnings

kalisjoshua:
Closes #115 

## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No


kalisjoshua:
This is just a suggestion to clean up the lint warnings. I am not passionate about this solution but I do like lint output to be clean.

belsrc:
I'm assuming they were just warnings, so not the end of the world. But I'm fine with sentiment. Overall though, I would prefer just targets ignore comments over an alias.

kalisjoshua:
Yeah, I am not a big fan either. I just was floating it out there because I was being lazy and didn't want to litter `// biome-ignore...` all over the place. I've had a setup in the past that would allow explicit `any` and disallow implicit `any`. Feelings?

belsrc:
> I've had a setup in the past that would allow explicit `any` and disallow implicit `any`.

Interesting thought. My feeling is that it could be "abused." In the sense that there is nothing stopping someone from `any`ing everything as long as it is explicit. Which is just more work for all parties involved.

brandonjpierce:
> Yeah, I am not a big fan either. I just was floating it out there because I was being lazy and didn't want to litter `// biome-ignore...` all over the place. I've had a setup in the past that would allow explicit `any` and disallow implicit `any`. Feelings?

I'm actually for this if we can make sure the various linters could all theoretically support it

kalisjoshua:
I'm fine with this approach till it doesn't work. I just don't like to see the warnings in the output.

kalisjoshua:
I can appreciate the desire for more eloquence as remote communication is always more difficult.

It sounds like you interpreted my comment as me wanting to not allow using `any` explicitly; this is not the case. I am fine with `any` in key areas and when necessary; these will always exist.

I think we are also aligned with where the "burden" of responsibility should reside; on the author not the reviewer. The author should be responsible for clearly describing the changes they are introducing both in implementation logic and type definition.

The use of `any` should be avoided as often as possible as it doesn't help with understanding what the shapes of values should be.

There are far more uses of `// biome-ignore ...` without a description filled out than with one filled out.

What I was suggesting is allow explicit `any` (without a warning) and disallow implicit `any` (as an error).

belsrc:
> What I was suggesting is allow explicit `any` (without a warning) and disallow implicit `any` (as an error).

That is exactly what I am referring to, though. When I said `any` above, I mean `explicit any` (with no warn).

> There are far more uses of // biome-ignore ... without a description filled out than with one filled out.

I would have to have you argue your case for that. As, in my eyes, its no different then `@ts-x <reason>`. Both should only be used when there is no other option. And their use should be justified. It is kind of a disservice to future maintainers (which may be oneself) if there is no reason provided. As no one may know/remember why something is there in the first place.

brandonjpierce:
I think the largest value add for explicit `@ts-expect-error` or `biome-ignore` is really for that additional description/reason context for _why_ the `any` is required. I've gone back through so many old codebases and just thought to myself "why the heck didn't I ever fix this?" only to slowly remember why 😆

kalisjoshua:
> I would have to have you argue your case ...

I am not making a case for anything, I am stating a fact. Search for `// biome-ignore lint/suspicious/noExplicitAny` across the code base on "main" and there are more instances without an explanation than there are with one. Broadening the search to just `// biome-ignore` and analyzing the results shows that there are more instances without than there are with; and of the ones that have something very few are actually explaining a reason for the exception.

I am not trying to be critical or getting on a high horse; because I am also guilty of not doing a good job of filling in an explanation myself. I am just pointing out that there does not seem to be rigor in requiring the explanation currently.

What I would argue is that disabling the warning for explicit use of `any` would be cleaner than forcing us to add the ignore comment for each instance. It would also be less work as disabling it would be effort once and keeping it as a warning requires ignoring continuously into the future.

belsrc:
Going in circles.

> What I would argue is that disabling the warning for explicit use of any would be cleaner than forcing us to add the ignore comment for each instance. It would also be less work as disabling it would be effort once and keeping it as a warning requires ignoring continuously into the future.

👉 https://github.com/gohypergiant/standard-toolkit/pull/116#issuecomment-2515436625

> I am not trying to be critical or getting on a high horse; because I am also guilty of not doing a good job of filling in an explanation myself. I am just pointing out that there does not seem to be rigor in requiring the explanation currently.

That is very much an "us" thing and in no way invalidates the point. I don't know where the `<explanation>` stuff is coming from (VSC crap?) but if you omit it, biome already tells you to do better.

Screenshot 2024-12-04 at 2 56 48 PM

kalisjoshua:
Ok good. Then I was just misunderstanding.

So we want to change the setting to allow for explicit `any` without an disable comment required?

brandonjpierce:
@kalisjoshua TL;DR:

- We keep the rule for no `any` (error)
- IF you need an `any`, we require an explicit `biome-ignore` comment including description to indicate the nuance of why we need it

This way, any `any` that exists in the codebase, should have an appropriate context attached to it vs. some may or may not having it with the explicit `any` rule

brandonjpierce:
@kalisjoshua Only additional change required before we merge is reversing the `any` lint rule back to `warn` and adding any required explicit ignore comments w/ context descriptions

kalisjoshua:
Updated @brandonjpierce

packages/core/src/composition/compose/index.ts
```diff
@@ -1,10 +1,23 @@
...
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
 /* eslint-disable @typescript-eslint/no-explicit-any */
```
brandonjpierce, line 13: @belsrc tagging you for final polish pass since most of this was written by yourself :D are we good remove the `eslint-disable` lines across the board?

packages/core/src/composition/compose/index.ts
```diff
@@ -1,10 +1,23 @@
...
 import type { UnaryFunction } from '@/types';
 
 // https://stackoverflow.com/questions/49310886/typing-compose-function-in-typescript-flow-compose#answer-73082627
 
 // If its a list of functions, last being Unary
 type ComposeParams<Fns> = Fns extends readonly [
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
```
brandonjpierce, line 20: @belsrc similarly, are you able to give some justification / explanation for these ignores?

ArrayKnight: Can't most of these just be converted to `unknown`s?

brandonjpierce: IIRC @belsrc mentioned some issues with using `unknown`

packages/core/src/composition/compose/index.ts
belsrc: Shit, meant to get back to this and completely forgot. Yes, since we don't use eslint 😆

packages/core/src/composition/compose/index.ts
belsrc: Yea, there were some that get a little cranky when swapping.

belsrc (dismissed):
Sorry about that @kalisjoshua, catch up hell and completely forgot. One small nit, not going to cry over it so change if you want, but semantically, `Cache` should be `Table` or something since it isn't really a cache. In the context that it is being used, anyway.

packages/core/src/composition/compose/index.ts
kalisjoshua: Will delete.

packages/core/src/composition/compose/index.ts
kalisjoshua: I can edit this one to be `unknown` and see how things go.

kalisjoshua:
I've renamed 'Cache' to 'Table' @belsrc. Thanks for the suggestion.

I've also cleaned up a lot of other lint errors/warnings. There are still a few in there but they will require more thought. The ones I fixed are mostly a common pattern; we are repeating ourselves a lot.

PR #125, opened 2024-12-04
refactor: toBoolean does more; narrower predicates

### LINKED ISSUE #122: Evolve `toBoolean` and `isTrue` (et.al) implementations
The set of operations for boolean checks needs to be more clearly defined and offer more comparisons than just "true" or "false" checks with different names.

1. "false-values": '', 0, '0', '0.000', false, 'false', undefined, null, NaN
    - '0.000' - any number of leading or trailing zeros
    - 'false' - any capitalizations or space-padding
2. `toBoolean` (converter) - will consider any non-false value to be true
3. `is???` (predicate) - will specifically expect a string of '???' to be one of:
    - false
    - no (false)
    - off (false)
    - on (true)
    - true
    - yes (true)

**is??? truth table**

is??? | boolean
----- | -------
false | false
no    | false
off   | false
on    | true
true  | true
yes   | true

These changes will allow for loose checking (converters/to-boolean) where a value is simply coerced into a boolean while the more specific checking (predicates/is-noyes) will validate more specific values.

kalisjoshua:
the toBoolean function centralizes the logic for coercing a value and allows the predicate functions to be more specific in what they validate rather than them simply being alais names to broad validation.

Closes #122

## ✅ Pull Request Checklist:

- [x] Yes
- [ ] No

The function `toBoolean` no longer will consider specific values - 'no', 'off', 'n' to be "false" or 'on', 'yes', 'y' to be "true". If that is being relied on that assumption will break; no tests show this to be extant.

### DIFF
--- a/packages/converters/src/to-boolean/index.ts
+++ b/packages/converters/src/to-boolean/index.ts
@@ -10,30 +10,29 @@
  * governing permissions and limitations under the License.
  */
 
-import { isTrue } from '@accelint/predicates';
-
 /**
- * Compare the given value against a custom list of `truthy` values.
+ * Returns true for any value not found to be a "false" value.
  *
- * String values are not case sensitive.
+ * **"false" values**
+ *   - inherently false values: '' (empty string), 0, false, undefined, null, NaN
+ *   - numeric zero: '0.000' - any number of leading or trailing zeros
+ *   - string literal: 'false' - any capitalizations or space-padding
  *
- * _1, '1', 'on', 'true', 'yes', true_
+ * For more restrictive comparisons against: true, false, on, off, yes, no; see
+ * the predicates package (\@accelint/predicates).
  *
  * @pure
  *
  * @example
- * toBoolean('on');
- * // true
- *
- * toBoolean('yes');
- * // true
- *
- * toBoolean('off');
- * // false
- *
- * toBoolean('no');
- * // false
+ * toBoolean(1);          // true
+ * toBoolean(' FaLsE ');  // false
+ * toBoolean('  true');   // true
+ * toBoolean('000.000');  // false
  */
 export function toBoolean(val: unknown) {
-  return isTrue(val);
+  return !(
+    !val ||
+    `${val}`.trim().toLowerCase() === 'false' ||
+    Number.parseFloat(`${val}`) === 0
+  );
 }
--- a/packages/predicates/src/is-noyes/index.test.ts
+++ b/packages/predicates/src/is-noyes/index.test.ts
@@ -10,26 +10,43 @@
  * governing permissions and limitations under the License.
  */
 
-import { describe, it, expect } from 'vitest';
-import { isTrue, isYes, isFalse, isNo, isOn, isOff } from './';
+import { describe, expect, it } from 'vitest';
+import {
+  isAnyFalsy,
+  isAnyTruthy,
+  isFalse,
+  isNo,
+  isOff,
+  isOn,
+  isTrue,
+  isYes,
+} from './';
 
-const truthy = [1, '1', 'on', 'true', 'yes', true, 'ON', 'YES', 'TRUE'];
-const falsey = [0, '0', 'off', 'false', 'no', false, 'OFF', 'NO', 'FALSE'];
+type Config = {
+  negative: unknown[];
+  positive: unknown[];
+  predicate: (a: unknown) => boolean;
+};
 
-describe('boolean validators', () => {
-  for (const item of truthy) {
-    it(`should return true for ${item}`, () => {
-      expect(isOn(item)).toBeTruthy();
-      expect(isTrue(item)).toBeTruthy();
-      expect(isYes(item)).toBeTruthy();
+const falsy = [0, '', false, ' false ', null, undefined, Number.NaN];
+const truthy = [1, true, ' true '];
+
+describe('boolean predicates', () => {
+  describe.each`
+    predicate      | positive                              | negative
+    ${isAnyFalsy}  | ${[...falsy, 'no', 'off']}            | ${[...truthy, 'on', 'yes']}
+    ${isAnyTruthy} | ${[...truthy, 'on', 'yes']}           | ${[...falsy, 'no', 'off']}
+    ${isFalse}     | ${[...falsy, ' false', '0']}          | ${[...truthy, 'o', 'O', 'true', 'string with false', '0.00']}
+    ${isNo}        | ${[...falsy, ' n', 'N ', 'no', 'NO']} | ${[...truthy, 'yes', 'string with no']}
+    ${isOff}       | ${[...falsy, ' off', 'OFF ']}         | ${[...truthy, 'on', 'string with off', 'of']}
+    ${isOn}        | ${[...truthy, ' on', 'ON ']}          | ${[...falsy, 'of', 'string with on', 'o']}
+    ${isTrue}      | ${[...truthy, ' true']}               | ${[...falsy, 'any string', {}, [], /abc/]}
+    ${isYes}       | ${[...truthy, ' yes', 'YeS ', 'y']}   | ${[...falsy, 'no', 'string with yes', 2]}
+  `('$predicate.name', ({ predicate, ...lists }: Config) => {
+    describe.each(['positive', 'negative'])('%s matches', (list) => {
+      it.each(lists[list] as unknown[])('%j', (val) => {
+        expect(predicate(val)).toBe(list === 'positive');
+      });
     });
-  }
-
-  for (const item of falsey) {
-    it(`should return false for ${item}`, () => {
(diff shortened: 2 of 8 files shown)


kalisjoshua:
I was waiting to update the bench tests till after we agree on a general direction; no need to benchmark something that isn't going to move forward.

packages/converters/src/to-boolean/index.ts
```diff
@@ -10,30 +10,33 @@
...
+ *
+ * toBoolean('000.000');
  * // false
  */
 export function toBoolean(val: unknown) {
-  return isTrue(val);
+  return !(
+    !val ||
+    `${val}`.trim().toLowerCase() === 'false' ||
+    Number.parseFloat(val as string) === 0
```
ArrayKnight, line 40: Why not wrap it in a string template literal? Since `unknown` is definitely not `string`

packages/predicates/src/is-noyes/index.ts
```diff
@@ -87,39 +79,42 @@ export const isNo = isFalse;
  * // true
  *
  * isOff('no');
+ * // false
  */
-export const isOff = isFalse;
+export const isOff = (val: unknown) => !val || test(['off'], val);
```
ArrayKnight, line 84: There's a little bit of a discrepancy between `isOff` and `isOn`, where `isOff` will accept any falsy value || `off` variants, instead of just `false` like how `isOn` only accepts `true` || `on` variants

I would expect them to be the same level of rigid/flexible. And I'd expect them both to accept things like `0` or `1` since that's relatively common term for on/off

ArrayKnight:
Feels like maybe we also want something that is `isYes + isTrue + isOn` & `isNo + isFalse + isOff`, like `isTruthy` & `isFalsy`?

packages/predicates/src/is-noyes/index.ts
kalisjoshua: Good point.

packages/converters/src/to-boolean/index.ts
belsrc: I think he meant `Number.parseFloat(val as string)` -> ``Number.parseFloat(`{val}`)``

kalisjoshua: That is why I asked. :)

That makes more sense.

kalisjoshua: I can, and have changed it, but I remember my original reasoning being that `as string` is only for typescript so it will be removed in production while wrapping in a string template would carry into production.

belsrc: I would agree with that though. `Number.parseFloat({}) === NaN` which doesn't `=== 0` so the logic checks out. Which I guess a case could be made, are we considering `NaN` as `false`?

kalisjoshua: It is a good question. I asked myself that as well. What I fell back on is that Javascript considers is to be false so I felt that staying consistent with the language is best.

kalisjoshua:
> Feels like maybe we also want something that is `isYes + isTrue + isOn` & `isNo + isFalse + isOff`, like `isTruthy` & `isFalsy`?

The converter `toBoolean` is the most liberal and could be combined with the more specific calls to gain those benefits I think.

I don't think it is a good idea to include all of the specific cases - yes + true + no or no + off + false - into a single call.

kalisjoshua:
I decided to sever the relationship between `converters/toBoolean` and `predicates/is-noyes` because I think they are not as compatible as might appear.

First thought, for `toBoolean` there should only two possible return values: `true` or `false`; it should never throw an error or return null or undefined. Therefor it needs to be liberal in it's conversion and that leads to the ethos of, "anything, that is something not explicitly false, is true." The other thought I had along these lines is, "if there is a list for true and list for false and we check against those two lists, then anything not in those lists would be an alternate return value." All of this leads me to believe that there is a small set of "false" values and anything outside of that set is not false, or true; at least for the purposes and domain of this type of function.

Second, for the predicates, `toBoolean` becomes problematic for truthy values because it's liberal criteria conflicts with the more restrictive ethos of `is*` functions. For instance because `toBoolean('no')` will evaluate to `true` then `isYes('no')` would evaluate to `true` without a more complicated check than `['y', 'yes', 1, 'true].includes(val)`. Also, I felt that it would be better to be consistent within `is-noyes` function implementations as well as limit interdependencies where not prohibitive.

packages/converters/src/to-boolean/index.ts
```diff
@@ -10,30 +10,29 @@
...
- * toBoolean('no');
- * // false
+ * toBoolean(1);          // true
+ * toBoolean(' FaLsE ');  // false
+ * toBoolean('  true');   // true
+ * toBoolean('000.000');  // false
  */
 export function toBoolean(val: unknown) {
-  return isTrue(val);
+  return !(
```
brandonjpierce, line 33: Small nit for cleanliness (and any other long term checks we might do):

```
function notCheck(val) {
  return !val;
}

function stringCheck(val) {
  return string.trim().toLowerCase() === 'false';
}

function numberCheck(val) {
  return Number.parseFloat(`${val}`) === 0;
}

export function toBoolean(val) {
  return !(notCheck(val) || stringCheck(val) || numberCheck(val))
}
```

packages/predicates/src/is-noyes/index.ts
```diff
@@ -10,154 +10,148 @@
...
- * // true
- *
- * isFalse('no');
+ * isFalse('');        // true
+ * isFalse(0);         // true
+ * isFalse(1);         // false
+ * isFalse(true);      // false
  */
-export const isFalse = (val: unknown) => test(falseValues, val);
+export const isFalse = (val: unknown) => test(listFalse, val);
```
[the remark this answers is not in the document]
kalisjoshua: Is that a use case? Is a value likely to be "Off" or "No"?

In my experience I would expect a value to be either: "On" or "Off" or `true` or `false` but nothing else and another value might be of a different variant.

packages/converters/src/to-boolean/index.ts
kalisjoshua: Seems verbose for not much value and additional overhead - cognitive and execution - in my opinion. I would more often lean on comments for further documentation of why each exists than additional code to execute.

packages/predicates/src/is-noyes/index.ts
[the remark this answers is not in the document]
brandonjpierce: It is a use case for us on a project now, yeah, as we have seen both variations due to bad data and would like a single function to accommodate all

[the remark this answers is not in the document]
brandonjpierce: We can kick it to another ticket / PR though, I am good with merging as is

brandonjpierce:
@kalisjoshua will need a changeset

kalisjoshua:
Changeset added.

packages/predicates/src/is-noyes/index.ts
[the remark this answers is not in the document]
orteth01: i think I prefer `isFalsy`/`isTruthy`, but I understand the motivation behind the current naming and am not opposed to leaving it

packages/predicates/src/is-noyes/index.test.ts
```diff
@@ -10,26 +10,43 @@
...
+  describe.each`
+    predicate      | positive                              | negative
+    ${isAnyFalsy}  | ${[...falsy, 'no', 'off']}            | ${[...truthy, 'on', 'yes']}
+    ${isAnyTruthy} | ${[...truthy, 'on', 'yes']}           | ${[...falsy, 'no', 'off']}
+    ${isFalse}     | ${[...falsy, ' false', '0']}          | ${[...truthy, 'o', 'O', 'true', 'string with false', '0.00']}
+    ${isNo}        | ${[...falsy, ' n', 'N ', 'no', 'NO']} | ${[...truthy, 'yes', 'string with no']}
+    ${isOff}       | ${[...falsy, ' off', 'OFF ']}         | ${[...truthy, 'on', 'string with off', 'of']}
+    ${isOn}        | ${[...truthy, ' on', 'ON ']}          | ${[...falsy, 'of', 'string with on', 'o']}
+    ${isTrue}      | ${[...truthy, ' true']}               | ${[...falsy, 'any string', {}, [], /abc/]}
+    ${isYes}       | ${[...truthy, ' yes', 'YeS ', 'y']}   | ${[...falsy, 'no', 'string with yes', 2]}
```
orteth01, line 44: just curious: does `'YeS '` reflect a value we actually see in the real use cases?

brandonjpierce: YeS, and also yES, and also YES 😆

brandonjpierce (approved):

PR #127, opened 2024-12-05
ADR: Documentation Site Tools Selection

kalisjoshua:
This is an early POC and documentation of goals in an ADR for discussion.

### DIFF
--- a/documentation/adrs/001-documentation-tool.md
+++ b/documentation/adrs/001-documentation-tool.md
@@ -1,0 +1,139 @@
+<!-- Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at https://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License. -->
+
+# Choosing a Documentation Tool
+
+We want to offer exemplary DevTK documentation for current and future users to learn and discover the functionalities contained.
+
+## Goals
+
+  1. Docblocks inside the codebase
+      - These are inline comments documenting the structure, usage, and behavior of specific functions, classes, and modules.
+      - Docblocks are helpful for users and maintainers as they provide information in close to where the code is being used.
+  2. Markdown outside of the codebase
+      - Markdown files outside the codebase can contain additional documentation such as high-level architectural decisions, design intentions, and contextual explanations.
+  3. Easy maintenance
+      - There are at least two aspects to maintenance: maintaining the documentation to keep it accurate, and generating the tool used to browse the documentation.
+      - The Docblocks are helpful but including exhaustive documentation in the codebase will become burdensome and avoided.
+      - Maintaining two (or more) sources of truth for any content will mean that it will become out of date very easily and quickly; the only thing worse than no documentation is bad/incorrect documentation.
+      - We will have plenty of work building the modules we don't need the additional responsibility of keeping a whole other project - the documentation site - up to date and accurate.
+  4. "No-setup" playgrounds
+      - Where possible, for each module, there could be an interactive component that would allow users to experiment with the codebase from within the documentation site itself.
+
+## Recommendation
+
+Based on the maturity of the tool, the track record of success, the developer experience, and customizability the tool(s) we are moving forward with will be TypeDoc + Docusaurus. TypeDoc will render the content from docblocks in code and Docusaurus will allow for better content organization and plugin options.
+
+## Tools Considered
+
+Tool         | Latest    | Releases  | Contributors | Build | Setup | React | Customization | Disqualifier
+:----------- | --------- | --------- | ------------ | ----- | ----- | ----- | ------------: | ------------
+[Docusaurus] | 2 hours   | 148       | 1175         | Yes   | easy  | Yes   | Moderate      |
+[TypeDoc]    | 2 days    | 224       | 227          | Yes   | easy  | Yes   | Moderate      | Better options?
+_            |           |           | Disqualified |       |       |       |               | _
+[Docsify]    | 1 day     | 14        | 183          | None  | easy  | Ext   | Minimal       | Burden of extension
+[Docz]       | 2 years   | 127       | 134          | Yes   | att0  | Yes   | Moderate      | Age without update
+[FumaDocs]   | 2 hours   | 673*      | 43*          | Yes   | att0  | Yes   | Moderate      | One developer
+[JSDoc]      | 2 days    | 12        | 83           | Gen   | easy  | No    | Extensive     | Burden of extension
+[MkDocs]     | 1 month   | 15        | 244          | Yes   | att0  | Yes?  | Extensive     | Burden of extension
+[Nextra]     | 9 hours   | 619*      | 169          | Yes   | hard  | Yes   | Moderate      | Poor dev experience
+[Pandoc]     | 1 day     | 140       | 522          | Yes   | att0  | Ext   | Extensive     | Burden of extension
+[RSPress]    | 2 days    | 115       | 97           | Yes   | easy  | Yes   | Moderate      | Maintained by Bytedance
+[Slate]      | 10 months | 28        | 126          | Yes   | hard  | No?   | Extensive     | Burden of extension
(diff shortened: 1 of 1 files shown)


documentation/adrs/001-documentation-tool.md
```diff
@@ -0,0 +1,69 @@
...
+  3. Easy maintenance
+      - There are at least two aspects to maintenance: maintaining the documentation to keep it accurate, and generating the tool used to browse the documentation.
+      - The Docblocks are helpful but including exhaustive documentation in the codebase will become burdensome and avoided.
+      - Maintaining two (or more) sources of truth for any content will mean that it will become out of date very easily and quickly; the only thing worse than no documentation is bad/incorrect documentation.
+      - We will have plenty of work building the modules we don't need the additional responsibility of keeping a whole other project - the documentation site - up to date and accurate.
+  4. "No-setup" playgrounds
+      - Where possible, for each module, there could be an interactive component that would allow users to experiment with the codebase from within the documentation site itself.
+  5. Limits
+      - dont want to replace training suite
+      - dont want to replace code comments
```
switzerb, line 21: I made a comment in the chat about this but I think it is worth addressing audience as a part of the requirements.  It is a nice sanity check on decisions that get made to ensure that multiple perspectives and roles are getting addressed and can access/use/contribute to the docs. 

- Designers 

Making the systems more transparent and consolidated will allow designers to improve and integrate new designs and make efforts more efficient and targeted 

- Application Developers (Jeric2o, Hawkeye, etc) 
As the consumers of the platform resources, enabling easy access and clear documentation on how to leverage what is available will help developer velocity and adoption. 

- Platform Developers 
As builders of the platform, devs will have a clear channel for explaining how and why the resources should be used, reducing mental load on communication strategies 

- Onboarding Developers and Designers 
Onboarding people will be simpler because there will be a consolidated way to communicate a lot of context and the ability to reference important materials 

- Product Owners and Managers 
Access to understanding the up to date capabilities makes it easier to plan and integrate the tools that are available into the existing product systems 

- Business Stakeholders 
By having a clear overview of the capabilities and systems that are in usage, it is more possible to see the 30,000 foot view and conceptualize places to scale and grow

kalisjoshua: What are/were your thoughts around "worth making sure whatever is chosen works across the board even for non-coder types"? Do you mean the documentation content (end result) or more the editing and maintaining of the content?

kalisjoshua:
I've been working with RSPress for the past few days and I did get it to nearly be doing what we would want it to be. Specifically, it has built-in support for rendering React components but not Javascript/TypeScript "previews" (playgrounds) so I've been working on putting that together. I've successfully integrated the [Monaco Editor](https://microsoft.github.io/monaco-editor/), which is nice, but getting full support for the modules - meaning that false errors aren't represented - is going to be considerably more difficult; maybe that's why no one else has already done it. I was looking into RunKit but wasn't getting it working immediately, I might return to it; however, I do think that it requires that the modules to be used will need to be published which poses a problem for getting things working before publishing.

belsrc:
> I was looking into RunKit but wasn't getting it working immediately

Can you expand on that a little? It should be as be something like 

```html
<script src="https://embed.runkit.com" data-element-id="my-element"></script>

<div id="my-element">
var getJSON = require("async-get-json");
await getJSON("https://storage.googleapis.com/maps-devrel/google.json");
</div>
```

To get it spun up. Is it where in the pipeline to inject the snippet the problem or something else?

kalisjoshua:
> > I was looking into RunKit but wasn't getting it working immediately
> 
> Can you expand on that a little? It should be as be something like
> 
> ```
> <script src="https://embed.runkit.com" data-element-id="my-element"></script>
> 
> <div id="my-element">
> var getJSON = require("async-get-json");
> await getJSON("https://storage.googleapis.com/maps-devrel/google.json");
> </div>
> ```
> 
> To get it spun up. Is it where in the pipeline to inject the snippet the problem or something else?

When I was experimenting, adding the script tag wasn't working at all. It wasn't showing up in the rendered page and there was no network requests in the dev-tools requesting the resources from RunKit.

I also don't think we actually need RunKit specifically because everything we are building is not relying on Node.js APIs at all and therefor can be run in the browser just as successfully as it would anywhere else. Additionally, using RunKit would require that whatever we want to run in it be published and available which limits our ability to test unpublished modules.

documentation/adrs/001-documentation-tool.md
```diff
@@ -0,0 +1,107 @@
...
+- __Customization__ - the level needed to accomplish our goals
+  - Minimal
+  - Moderate
+  - Extensive
+- __Latest__ - most recent change in github (_at time of collection_)
+- __React__ - renders react components (e.g. from design system)
+  - Yes - native support
+  - Ext - will require an extension/plugin
+- __Releases__ - number of "releases" or production deployments
+- __Setup__ - ease of getting it running: easy, or hard
```
switzerb, line 70: One thing I don't see in any of the criteria is search or any kind of discoverability. One of the most difficult parts of docs is finding what you are looking for -- did RSPress have any kind of indexing at all? If we wanted that would we have to build our own plugins?

documentation/adrs/001-documentation-tool.md
switzerb: I was referring to contributions specifically. The technical bits of the docs are important, but ultimately the docs are only as good as the content...and a lot of the content will be "business" content that is owned by designers or potentially technical writers. But it looks like most of these use MDX or markdown text files so that seems doable.

documentation/adrs/001-documentation-tool.md
switzerb: One other requirement that seems reasonable is that whatever we chose, the content shouldn't be "locked" into the platform. Like with Sharepoint it is nearly impossible to move content out of sharepoint and into another context. Although I suppose if most of the writing is in markdown, then that should be open enough.

documentation/adrs/001-documentation-tool.md
kalisjoshua: Markdown (and similar i.e. MDX) was an unofficial requirement for me I think.

documentation/adrs/001-documentation-tool.md
kalisjoshua: Search is supported "out of the box". 
image

Great question! And of the "top" contenders search is supported equally I would say.

belsrc:
> When I was experimenting, adding the script tag wasn't working at all. It wasn't showing up in the rendered page and there was no network requests in the dev-tools requesting the resources from RunKit.

Interesting. I would need to take a look at that.


> I also don't think we actually need RunKit specifically because everything we are building is not relying on Node.js APIs at all and therefor can be run in the browser just as successfully as it would anywhere else. Additionally, using RunKit would require that whatever we want to run in it be published and available which limits our ability to test unpublished modules.

It's not really about Node APIs or not. Its more of having a REPL to interactively using the module in question. Like Lodash and, I believe, Rambda have. Unpublished part is not really a concern to me. If it is not published, no one can use it. No documentation needed. No REPL needed.

kalisjoshua:
> It's not really about Node APIs or not. Its more of having a REPL to interactively using the module in question.

We can a accomplish a REPL very simply without RunKit; I created a Javascript-only POC in a few hours including the specific lib dependencies. The challenge comes with including TypeScript features in an editor but even that is very accomplishable with something like StackBlitz and some additional config.

belsrc:
> I created a Javascript-only POC in a few hours including the specific lib dependencies.

Why add more maintenance overhead when something already exists?

kalisjoshua:
> Why add more maintenance overhead when something already exists?

Why add a dependency and potential security vector to avoid a very simple function?

belsrc:
I'm going to go out on limb and say "making an in page repl" is not just a "simple function". And OSS packages are usually more secure then one persons NIH implementation. Coming from a strict security oriented stand-point.

kalisjoshua:
We will use a dependency for the editor - monaco, stackblitz, etc. - because they are going to provide a whole lot of functionality for little investment. Using RunKit (or something like it) gives little functionality - and unnecessary limitations - for relatively more investment - than a local solution - and include a performance impact as well. I just don't see the return on investment for the REPL specifically.

As an example, here is the whole implementation of a "RE" of REPL where something like rspress provides the "PL".

```

```

ArrayKnight:
Before we get too far, I do want to resurface @shadyendless 's concern that Rspress is created by ByteDance, which might be a no-no considering our rules against TikTok. I know this is an open source dependency being used by an open source project, but worth double checking with legal

kalisjoshua:
> Before we get too far...

Who should reach out, and who to? I'm happy to I just need to know who to.

orteth01 (approved):

belsrc (approved):

ArrayKnight (approved):

PR #145, opened 2024-12-27
update tsconfig.json with audit script

kalisjoshua:
- Add a CLI utility for auditing and updating the tsconfig.json definition.
- Update the sample `toBoolean` doc page to:
    - Import the correct generated file
    - Clean a console error ('iterated items should have a "key" property')

### DIFF
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -1,30 +1,27 @@
 {
-  "$schema": "https://biomejs.dev/schemas/1.8.0/schema.json",
-  "organizeImports": {
-    "enabled": true
-  },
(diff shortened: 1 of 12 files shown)


kalisjoshua:
Anyone a `tsconfig` "wizard"? Including all of the packages in the config for the docs site is - I think? - necessary to access them; however, doing so is reporting a bunch of typescript errors that aren't actually errors.

image

Steps to reproduce

1. Checkout the branch
2. `pnpm --filter @accelint/docs dev`

kalisjoshua:
Going a different direction (hopefully).

PR #146, opened 2024-12-27
chore(docs): use typedoc directly

orteth01:
## some notes

- putting the generated assets from ladle and typedoc in /docs/src/public was the simplest way i could find to get everything to work in both dev mode and after building. stuff in that dir is just copied to dist [reference](https://rspress.dev/guide/basic/static-assets#other-static-assets)
- right now ladle build and typedoc run before both rspress dev/rspress build. it takes a while, which is kind of annoying. we can likely optimize this if it's prohibitive to development of the docs.

## testing
- do a fresh dep install

DEV
- pnpm --filter=@accelint/docs dev
- verify that the toBoolean doc looks correct
- verify the design-system component story embed works

BUILD/PREVIEW
- pnpm --filter=@accelint/docs build
- pnpm --filter=@accelint/docs preview
- verify that the toBoolean doc looks correct
- verify the design-system component story embed works

Screenshot 2024-12-27 at 3 55 44 PM

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "pnpm run build:deps && rspress dev",
     "build:deps": "pnpm run build:ladle && pnpm run build:typedoc",
-    "build:ladle": "pnpm --filter=@accelint/design-system exec ladle build -o ../../apps/docs/ladle --base ladle",
+    "build:ladle": "pnpm --filter=@accelint/design-system exec ladle build -o ../../apps/docs/src/public/ladle --base ladle",
     "build:typedoc": "typedoc",
     "build": "pnpm run build:deps && rspress build",
     "preview": "rspress preview",
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -13,6 +13,7 @@
     "preview": "rspress preview"
   },
   "dependencies": {
+    "@accelint/design-system": "workspace:*",
     "@rspress/plugin-playground": "^1.38.0",
     "@types/node": "^18.19.67",
     "react": "^18.3.1",
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -3,18 +3,22 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "build": "rspress build && pnpm run build-ladle",
-    "build-ladle": "pnpm --filter=@accelint/design-system exec ladle build -o ../../apps/docs/dist/ladle --base ladle",
+    "dev": "pnpm run build:deps && rspress dev",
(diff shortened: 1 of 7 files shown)


apps/docs/package.json
```diff
@@ -3,19 +3,22 @@
...
   "private": true,
   "scripts": {
-    "build": "rspress build && pnpm run build-ladle",
-    "build-ladle": "pnpm --filter=@accelint/design-system exec ladle build -o ../../apps/docs/dist/ladle --base ladle",
+    "dev": "pnpm run build:deps && rspress dev",
+    "build:deps": "pnpm run build:ladle && pnpm run build:typedoc",
+    "build:ladle": "pnpm --filter=@accelint/design-system exec ladle build -o ../../apps/docs/src/public/ladle --base ladle",
+    "build:typedoc": "typedoc",
+    "build": "pnpm run build:deps && rspress build",
+    "preview": "rspress preview",
```
kalisjoshua, line 11: Not a requirement, just a preference of mine that can be ignored, I like to order things in alpha-order. There is a command in VSCode to do this for you if, like me, you don't want to do tedious things like this.

apps/docs/src/packages/converters/to-boolean.mdx
```diff
@@ -1,5 +1,5 @@
 import Install from '/lib/install';
-import Typedoc from "/src/typedocs-output/functions/toBoolean.md";
+import Typedoc from "/src/public/typedoc/@accelint/converters/functions/toBoolean.md";
```
kalisjoshua, line 2: I like this output path more than the plugin.

apps/docs/package.json
orteth01: [updated](https://github.com/gohypergiant/standard-toolkit/pull/146/commits/9bee35cf58d74a45cb0240a9d87bf8ea7f39162c)

apps/docs/package.json
```diff
@@ -3,19 +3,23 @@
...
+    "build:typedoc": "typedoc",
+    "build": "pnpm run build:deps && rspress build",
     "check": "biome check --write",
-    "dev": "rspress dev",
+    "dev": "pnpm run build:deps && rspress dev",
     "format": "biome format --write",
     "preview": "rspress preview"
   },
   "dependencies": {
+    "@accelint/design-system": "workspace:*",
```
orteth01, line 16: because the docs build step depends on the design system being built (for ladle), we need to specify it as a dep here. this way, turbo knows that we can only build the docs after the design system has been built. without this, we see build errors like [this](https://github.com/gohypergiant/standard-toolkit/actions/runs/12599681591/job/35117117576?pr=146)

PR #156, opened 2025-01-06
fix(lint): tsdoc warnings cleanup

kalisjoshua:
Cleanup of tsdoc warnings and tag usage.


apps/docs/typedoc.config.mjs
```diff
@@ -12,14 +12,19 @@
...
 export default {
-  entryPointStrategy: 'packages',
+  customCss: './custom.css',
+  disableSources: true,
   entryPoints: ['../../packages/*'],
+  entryPointStrategy: 'packages',
+  githubPages: false, // false === do not emit the file `.nojekyll`
+  name: 'Accelint Development Toolkit (DevTK)',
+  navigationLinks: {
+    // biome-ignore lint/style/useNamingConvention: <explanation>
```
orteth01, line 22: we could probably ignore naming convention for config files. if we leave the comment, let's add a real explanation

packages/converters/src/boolean-to-number/index.ts
```diff
@@ -13,7 +13,8 @@
 /**
  * Convert a boolean to an integer
  *
- * @pure
+ * @remarks
+ * pure function
```
orteth01, line 17: 👍

packages/core/src/combinators/a/index.ts
```diff
@@ -15,7 +27,5 @@ export const A =
     (x: T) =>
       f(x);
 
-/**
- * @alias A
- */
+/** {@inheritDoc A} */
```
orteth01, line 30: cool

orteth01 (approved):

apps/docs/typedoc.config.mjs
kalisjoshua: I couldn't get Biome to ignore this so I put a small description.

PR #157, opened 2025-01-06
chore: embed ladle in typedoc

orteth01:
Screenshot 2025-01-06 at 11 58 13 AM

also updated the link


apps/docs/typedoc.config.mjs
```diff
@@ -16,6 +16,7 @@ export default {
   entryPoints: ['../../packages/*'],
   packageOptions: {
     entryPoints: ['src/index.ts'],
+    projectDocuments: ['src/**/*.md'],
```
orteth01, line 19: note: projectDocuments needs to be within packageOptions because we're using the "packages" entryPointStrategy

kalisjoshua (approved):

PR #158, opened 2025-01-06
chore: enable lint of design-system package

### LINKED ISSUE #148: Enable lint for design system package a

orteth01:
Closes #148 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

- design system `lint` should be part of CI checks
- CI lint and test checks should pass

- [ ] Yes
- [x] No

## 💬 Other information

Some errors I suppressed with a comment. there were a few reasons:
- intentional
- larger refactor required

### DIFF
--- a/packages/design-system/src/components/query-builder/group.tsx
+++ b/packages/design-system/src/components/query-builder/group.tsx
@@ -1,3 +1,15 @@
+/*
+ * Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
 import { Fragment, useContext, useMemo } from 'react';
 import {
   type RuleGroupArray,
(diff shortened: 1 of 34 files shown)


packages/design-system/src/hooks/use-defaults/types.ts
```diff
@@ -57,43 +69,81 @@ type DefaultsOf<T extends Record<string, object>> = {
 
 // TODO: Figure out a way to nest related components(?)
 export type DefaultsContext = DefaultsOf<{
+  // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
```
orteth01, line 72: this is.... very annoying 😅

it would be nice if we could do ignore blocks. i'm hesitant to ignore the entire file.

orteth01:
Screenshot 2025-01-06 at 1 53 14 PM

confirmed this ran on this PR

packages/design-system/src/components/query-builder/group.tsx
```diff
@@ -229,27 +241,29 @@ export function RuleGroupHeaderComponents(
...
-              context={ruleGroup.context}
-              validation={ruleGroup.validationResult}
-              schema={ruleGroup.schema}
-            />
-          </div>
-        )}
+      {!(
+        ruleGroup.schema.showCombinatorsBetweenRules ||
+        ruleGroup.schema.independentCombinators
+      ) && (
```
orteth01, line 247: i don't like this change. i feel like the `!` is sneaky and can be glossed over easily. if other folks agree, i'm happy to revert the change and ignore the lint error instead

kalisjoshua: That isn't the highest priority in that component (for me) for refactor. I'm fine with it this way; I don't have a strong feeling either way in that file as it is.

packages/design-system/src/utils/props.ts
```diff
@@ -62,6 +74,7 @@ function mergeStaticClassName<T extends RenderProps<object>>(
         return acc;
       }
 
+      // biome-ignore lint/performance/noAccumulatingSpread: we can discuss refactoring separately
```
belsrc, line 77: Wouldn't this one just be ... ?
```ts
acc.push(props.className);

return acc;
```

with the one below being

```ts
Object.assign(acc, props.style);

return acc;
```

Since they are in reducers and you don't need to worry about purity of the objects since they are created at the time of the `reduce` call.

packages/design-system/src/hooks/use-defaults/types.ts
belsrc: Agreed. Unfortunately Biome as repeatedly said they don't plan to allow ignoring a rule for a file. Just the entire file for everything. RIP ☠️

packages/design-system/src/utils/props.ts
belsrc: Yea. In this case since it is a reduce and only the accumulator is being mutated, it's fine.

orteth01: updated here: https://github.com/gohypergiant/standard-toolkit/pull/158/commits/a12e3640a8407bbbc21c775c58fc80516390548f

belsrc (approved):
Nothing else really jumped out at me (ignoring Biome's ignore decision).

PR #164, opened 2025-01-13
feat: Slider component

raechelo:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:
open the ladle page and view the Slider component. There are a handful of stories for review, including:
- controlled range slider
- controlled single value slider
- uncontrolled range slier
- uncontrolled single value slider
- slider with output (RAC components indicate an output value, but our designs indicate an _input_ value. The above components feature the input, but that inclusion can be toggled in the props menu by toggling `includeTextField` and `includeOutputField`

- [ ] Yes
- [x] No

## 💬 Other information
I didn't have a ticket for this, because I don't have Jira access yet, but here is the screenshot I was given and attempted to match.
image


packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,312 @@
+import {
```
raechelo, line 1: i have a feeling i didn't organize this file properly, please lmk if there is a better way to do this!

packages/design-system/.ladle/theme/slider.css.ts
```diff
@@ -0,0 +1,145 @@
+import { style } from '@vanilla-extract/css';
```
raechelo, line 1: I eyeballed all the styles from the pic, so I doubt they're fully accurate. If anything should be tweaked, I am happy to do so!

packages/design-system/src/components/slider/slider.stories.tsx
```diff
@@ -0,0 +1,490 @@
...
+    <Slider {...rest}>
+      {({ state }) => (
+        <>
+          {includeRangeLabel && <AriaLabel>{label}</AriaLabel>}
+          {includeOutputField && <AriaText>{state.values.join(',')}</AriaText>}
+          <SliderTrack>
+            {({ state }: SliderRenderProps) => (
+              <>
+                <div
+                  className={sliderBar}
```
raechelo, line 167: i think i applied this incorrectly, but was unable to get it working with the same method the `Picker` stories were using. User error, i'm sure 😆

packages/design-system/src/components/slider/slider.css.ts
```diff
@@ -0,0 +1,167 @@
...
+            query: { alignLabel: 'left' },
+            margin: 'auto 0',
+          }),
+        },
+      },
+    }),
+  },
+  thumb: {
+    container: style({
+      ' input': {
```
raechelo, line 129: looks like this is the line that is causing the build to fail. I wasn't sure how to properly address this issue, because the RAC component has an grandchild element of `input`, which holds a 90px width. The tooltip in the story is hovering over the `input` rather than the main level component, thus the width change. Below are a few pictures to better explain.

Screenshot 2025-01-13 at 2 48 00 PM
Screenshot 2025-01-13 at 2 47 05 PM
Screenshot 2025-01-13 at 2 46 51 PM

raechelo: fwiw, this is how RAC implements the component. You can see the same `input` width in this screenshot:

Screenshot 2025-01-13 at 2 54 40 PM

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+export type SliderClassNames = PartialDeep<{
+  slider: {
+    container: string;
+    label: string;
+    min: string;
+    max: string;
+  };
+  track: {
+    container: string;
+    track: string; // TODO: adjust the styles to use this class
```
raechelo, line 25: i have the TODO here, but i'm not quite sure i actually need to make that change 🤔💭

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+    [theme.Slider, classNamesProp],
+  );
+
+  const style = useCallback(
+    (renderProps: SliderRenderProps) =>
+      inlineVars(sliderStateVars, renderProps),
+    [],
+  );
+
+  const values = useMemo<
```
ArrayKnight, line 70: You shouldn't be providing the context that you're consuming. You can remove this

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+        typeof SliderOutputContext,
+        ContextValue<SliderOutputProps, HTMLOutputElement>,
+      ],
+    ]
+  >(() => [[SliderOutputContext, { classNames }]], [classNames]);
+
+  const children = useCallback(
+    (renderProps: SliderRenderProps) => (
+      <Provider values={values}>
+        {callRenderProps(childrenProp, renderProps)}
```
ArrayKnight, line 82: You seem to have a container, but not a styleable element inside of the container

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+        typeof SliderThumbContext,
+        ContextValue<SliderThumbProps, HTMLDivElement>,
+      ],
+    ]
+  >(() => [[SliderThumbContext, { classNames }]], [classNames]);
+
+  const children = useCallback(
+    (renderProps: SliderTrackRenderProps) => (
+      <Provider values={values}>
+        {callRenderProps(childrenProp, renderProps)}
```
ArrayKnight, line 142: Same here, RE: styleable element

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+      ref={ref}
+      className={classNames?.track?.container}
+      style={style}
+    >
+      {children}
+    </RACSliderTrack>
+  );
+});
+
+/**
```
ArrayKnight, line 160: Let's move this down so it's directly next to the SliderThumb component so it gets picked up by intellisense

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+    [[typeof AriaLabelContext, ContextValue<LabelProps, HTMLLabelElement>]]
+  >(() => [[AriaLabelContext, {}]], []);
+
+  const children = useCallback(
+    (renderProps: SliderThumbRenderProps) => (
+      <Provider values={values}>
+        <div className={classNames?.thumb?.thumb}>
+          {callRenderProps(childrenProp, {
+            ...renderProps,
+            orientation: 'horizontal',
```
ArrayKnight, line 200: Because the RAC supports a vertical orientation, ours should too

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+) {
+  [props, ref] = useContextProps(props, ref, SliderContext);
+
+  props = useDefaultProps(props, 'Slider');
+
+  const {
+    children: childrenProp,
+    classNames: classNamesProp,
+    alignLabel = 'top',
+    minValue = 0,
```
ArrayKnight, line 235: It looks like this is unused and because it's destructured it's not being passed along to the RAC

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+    thumb: string;
+  };
+  input: InputClassNames;
+  output: {
+    container: string;
+  };
+  group: GroupClassNames;
+}>;
+
+export type LabelAlignment = 'top' | 'left';
```
ArrayKnight, line 38: Because of the addition of supporting the vertical orientation, I think you're going to want to change this to being something like `'stacked' | 'inline'`

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+export type SliderRenderProps = RACSliderRenderProps;
+export type SliderThumbRenderProps = RACSliderThumbRenderProps;
+export type SliderTrackRenderProps = RACSliderTrackRenderProps;
+
+type BaseSliderProps = {
+  children?: RenderPropsChildren<SliderRenderProps>;
+  label?: string;
+  classNames?: SliderClassNames;
+  alignLabel?: LabelAlignment;
+  includeTextField?: boolean;
```
ArrayKnight, line 49: We shouldn't be having props for the `include*`. The determining factor for whether an element shows up or not is whether it was composed in or left out

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export type LabelAlignment = 'top' | 'left';
+
+export type SliderRenderProps = RACSliderRenderProps;
+export type SliderThumbRenderProps = RACSliderThumbRenderProps;
+export type SliderTrackRenderProps = RACSliderTrackRenderProps;
+
+type BaseSliderProps = {
+  children?: RenderPropsChildren<SliderRenderProps>;
+  label?: string;
```
ArrayKnight, line 46: This will be composed content, not passed through the parent

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  input: InputClassNames;
+  output: {
+    container: string;
+  };
+  group: GroupClassNames;
+}>;
+
+export type LabelAlignment = 'top' | 'left';
+
+export type SliderRenderProps = RACSliderRenderProps;
```
ArrayKnight, line 40: You'll likely want to wrap these with `AsType`

packages/design-system/src/components/slider/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  Required<Pick<BaseSliderProps, 'alignLabel'>>;
+
+export type SliderProps = Omit<
+  RACSliderProps,
+  'children' | 'classname' | 'style'
+> &
+  BaseSliderProps;
+
+type BaseSliderThumbProps = {
+  children?: RenderPropsChildren<SliderRenderProps>;
```
ArrayKnight, line 66: I don't think you need to redefine `children` here unless you're actually changing the type (which I don't think you are). And then you could simplify why getting rid of the BaseSlideThumbProps and just add the classNames prop directly to the join

ArrayKnight: This applies below too

packages/design-system/src/hooks/use-defaults/types.ts
```diff
@@ -122,6 +129,18 @@ export type DefaultsContext = DefaultsOf<{
   // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
   Select: SelectProps<object>;
   // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
+  Slider: SliderProps;
+  // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
```
ArrayKnight, line 133: Dear lord, who did this Biome commenting? Can you create an issue to convert to range suppression https://github.com/biomejs/biome/issues/4569 please?

packages/design-system/src/hooks/use-theme/types.ts
```diff
@@ -86,6 +87,8 @@ export type ThemeContext = {
   // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
   Select?: SelectClassNames;
   // biome-ignore lint/style/useNamingConvention: component name should be PascalCase
+  Slider?: SliderClassNames;
```
ArrayKnight, line 90: Here too

packages/design-system/src/components/slider/slider.stories.tsx
```diff
@@ -0,0 +1,490 @@
...
+                  style={{
+                    position: 'absolute',
+                    left: state.getThumbValue(1)
+                      ? `${state.getThumbValue(0)}%`
+                      : 0,
+                    width: state.getThumbValue(1)
+                      ? `${(state.getThumbValue(1) || 0) - (state.getThumbValue(0) || 0)}%`
+                      : `${state.getThumbValue(0)}%`,
+                  }}
+                />
```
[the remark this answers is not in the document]
ArrayKnight: Yea, let's go over this when we have time

ArrayKnight (changes requested):

packages/design-system/src/components/slider/slider.stories.css.ts
```diff
@@ -0,0 +1,7 @@
+import { style } from '@vanilla-extract/css';
+import { defaultSizesVarValues, semanticColorVars } from '../../styles';
+
+export const sliderBar = style({
+  backgroundColor: semanticColorVars.background.highlight.bold,
```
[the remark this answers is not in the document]
raechelo: for story only styling purposes. removed since we made the bar into a component

packages/design-system/src/components/slider/slider.tsx
raechelo: Since we don't have a design for that, how do you think i should handle the vertical style?

packages/design-system/src/hooks/use-defaults/types.ts
orteth01: that was me!

see PR/discussion here: https://github.com/gohypergiant/standard-toolkit/pull/158#discussion_r1904570169

we can create an issue to use the range suppression, but it looks like [it won't be available until 2.0 is released](https://github.com/biomejs/biome/pull/4649#issuecomment-2524003559)

packages/design-system/src/components/slider/slider.stories.tsx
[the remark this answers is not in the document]
raechelo: to to follow up for any possible future 👀 , this div was moved into a component of itself

packages/design-system/src/components/slider/slider.stories.tsx
raechelo: removed and added to the regular `slider.css.ts` file since this div was turned into a component

packages/design-system/src/components/slider/slider.css.ts
```diff
@@ -148,6 +164,8 @@ export const sliderClassNames: SliderClassNames = {
       height: sliderSpaceVars.track.height,
       margin: sliderSpaceVars.track.margin,
       right: 'auto',
+      marginInlineStart: 'calc(calc(12px / 2)* -1)',
+      marginInlineEnd: 'calc(calc(12px / 2)* -1)', // causes a clipping issue
```
raechelo, line 168: the `marginInlineEnd` doesn't exist in the Spectrum solution for this. While it works, it does leave a small clipping issue so it's not ideal. The Spectrum solution leverages a second track (which would be our third), and both are positioned based on the thumb's current location.

packages/design-system/src/components/slider/slider.css.ts
```diff
@@ -140,6 +140,22 @@ export const sliderClassNames: SliderClassNames = {
...
+          display: 'inline-block',
+          width: 'calc(100% - calc(calc(12px / 2) ) * 2)',
+          ':after': {
+            right: 'calc(calc(12px / 2) * -1)',
+            content: '',
+            inlineSize: 'calc(12px / 2)',
+            blockSize: '100%',
+            display: 'block',
+            position: 'absolute',
+          },
```
raechelo, line 158: should we decide to move forward with this implementation, all hardcoded sizes will be exchanged for variables

packages/design-system/src/components/slider/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+export {
+  Slider,
+  SliderBar,
+  SliderContext,
```
switzerb, line 18: I think you are missing a few exports from the files -- `SliderBarContext`, `SliderTrackContext`, `SliderThumbContext` -- so may want to doublecheck all the exports from all the files to make sure you caught everything.

packages/design-system/src/components/slider/slider.stories.tsx
```diff
@@ -0,0 +1,427 @@
...
+          )}
+        </SliderTrack>
+        <AriaText slot='min'>{minValue || 0}</AriaText>
+        <AriaText slot='max'>{maxValue || 100}</AriaText>
+      </>
+    )}
+  </Slider>
+);
+
+SliderExample.storyName = 'Slider with input';
```
switzerb, line 159: When the `includeOutputField` is on for this story, it comes in a little strangely with a visual overlap here:

CleanShot 2025-03-03 at 10 12 30

packages/design-system/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,415 @@
...
+
+  return (
+    <div
+      {...rest}
+      ref={ref}
+      className={classNames?.track?.bar}
+      style={{
+        position: 'absolute',
+        ...(state.orientation === 'horizontal' && {
+          left: state.getThumbValue(1) ? `${state.getThumbValue(0)}%` : 0,
```
switzerb, line 95: You should define these once outside of the inline style and then just reference them:

```
  const max = state.getThumbValue(1);
  const min = state.getThumbValue(0);

  return (
    <div
      {...rest}
      ref={ref}
      className={classNames?.track?.bar}
      style={{
        position: 'absolute',
        ...(state.orientation === 'horizontal' && {
          left: max ? `${min}%` : 0,
          width: max ? `${(max || 0) - (min || 0)}%` : `${min}%`,
        }),
        ...(state.orientation === 'vertical' && {
          bottom: max ? `${min}%` : 0,
          height: max ? `${(max || 0) - (min || 0)}%` : `${min}%`,
        }),
      }}
    />
  );
```

packages/design-system/src/components/slider/slider.stories.tsx
raechelo: > When the includeOutputField is on for this story, it comes in a little strangely with a visual overlap here:

oh good eyes. 🙈 what is the ideal solution here? do we expect developers to know better than to use the two together? should i just turn off the option to add `includeOutputField` label for this story?

apps/next/src/theme/slider.css.ts
```diff
@@ -0,0 +1,116 @@
...
+          }),
+        },
+      ]),
+    ),
+  },
+  track: {
+    container: style(
+      applyThemeVars<SliderState>(sliderStateVars, [
+        {
+          vars: assignPartialVars(sliderSpaceVars, {}),
```
switzerb, line 100: Why have an empty assignment here?

apps/next/src/theme/slider.css.ts
```diff
@@ -0,0 +1,116 @@
...
+      applyThemeVars<SliderThumbState>(sliderThumbStateVars, [
+        {
+          vars: assignPartialVars(sliderColorVars, {
+            background: semanticColorVars.background.highlight.bold,
+          }),
+        },
+        {
+          query: { isHovered: true },
+          vars: assignPartialVars(sliderColorVars, {
+            boxShadow: `0 0 0 6px rgb(from ${semanticColorVars.background.highlight.bold} r g b / 0.25)`,
```
switzerb, line 90: You might be able to pull this from `elevationVars`: https://github.com/gohypergiant/standard-toolkit/blob/main/packages/design-system/src/styles/theme.css.ts#L331

packages/design-system/.ladle/theme/slider.css.ts
```diff
@@ -0,0 +1,116 @@
...
+      applyThemeVars<SliderThumbState>(sliderThumbStateVars, [
+        {
+          vars: assignPartialVars(sliderColorVars, {
+            background: semanticColorVars.background.highlight.bold,
+          }),
+        },
+        {
+          query: { isHovered: true },
+          vars: assignPartialVars(sliderColorVars, {
+            boxShadow: `0 0 0 6px rgb(from ${semanticColorVars.background.highlight.bold} r g b / 0.25)`,
```
switzerb, line 90: same comment as above with the elevation vars

packages/design-system/.ladle/theme/slider.css.ts
```diff
@@ -0,0 +1,116 @@
...
+          }),
+        },
+      ]),
+    ),
+  },
+  track: {
+    container: style(
+      applyThemeVars<SliderState>(sliderStateVars, [
+        {
+          vars: assignPartialVars(sliderSpaceVars, {}),
```
switzerb, line 100: why the empty assignment?

packages/design-system/src/components/slider/slider.stories.tsx
switzerb: We might be able to use Typescript here to say "if you have one, you can't have the other. but yeah, for this I would just not have it as an option since a dev can include it if they want to in a different place.

switzerb (approved):

PR #166, opened 2025-01-14
colors alignment and theme from hypergiant.com

kalisjoshua:
attempted to use colors and stylistic themes from the corp webpage

image

image

### DIFF
--- a/apps/docs/src/css/custom.css
+++ b/apps/docs/src/css/custom.css
@@ -18,29 +18,114 @@
 
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-color-primary: #2e8555;
-  --ifm-color-primary-dark: #29784c;
-  --ifm-color-primary-darker: #277148;
-  --ifm-color-primary-darkest: #205d3b;
-  --ifm-color-primary-light: #33925d;
-  --ifm-color-primary-lighter: #359962;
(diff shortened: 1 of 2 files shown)


belsrc:
I honestly could not tell you if they would want corpo colors or project colors.

kalisjoshua:
I've reached out to Patrick G. The one significant improvement that is included here is the columns for lists because some of the modules/packages have a lot of elements an the columns are really helpful. Should I strip out everything else to move this forward? We can always overwrite these changes in the future, this was a good step into understanding how to customize the styles.

apps/docs/src/css/custom.css
orteth01, line 1: it's nice that this was so straightforward to modify

kalisjoshua:
We are good with approval for this design for now. No changes needed. @belsrc

orteth01 (approved):

belsrc (approved):
Cool. I am fine with this as a first pass if PG signed off on it. There are a couple of little oddities (like the search in dark mode looks off to me with the BG that it is on) but those can be addressed later.

kalisjoshua:
Agreed, it is not an end state for the styles by a long shot. And, yes that color was something I specifically ignored for now.

PR #167, opened 2025-01-14
refactor: Switch from ladle to storybook

### LINKED ISSUE #160: Component Playground - Switch from Ladle to Storybook
based on #137 as well as a few offline discussions, we have decided to switch to storybook for our component playground.

orteth01:
Closes #160 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

- fresh dep install (and build!)
- `pnpm --filter=@accelint/design-system run preview`
- click around, have a good time. change things, let the HMR flow.

- [ ] Yes
- [x] No

## 💬 Other information

I essentially just converted our existing stories to their storybook version. I think any more improvements or additions should be follow up work to prevent the scope of this PR from ballooning. 

we could potentially explore storybook features like these in the future:
- https://storybook.js.org/docs/api/doc-blocks/doc-block-colorpalette
- https://storybook.js.org/docs/api/doc-blocks/doc-block-typeset

> [!NOTE]
> HMR was very borked in Storybook 8. i spent a while trying to finagle it, to no avail. Storybook 7 HMR worked with no hassle, so we are using that version


packages/design-system/src/storybook/index.ts
orteth01, line 1: do we actually need this file, or can i just import directly from storybook/actions.ts where needed?

packages/design-system/.storybook/components.tsx
```diff
@@ -1,6 +1,20 @@
...
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// biome-ignore lint/correctness/noUnusedImports: fix later
+import React from 'react';
```
ArrayKnight, line 14: Why is this import necessary?

packages/design-system/.storybook/components.tsx
```diff
@@ -1,6 +1,20 @@
...
 import { clsx } from 'clsx';
 import { type PropsWithChildren, useMemo } from 'react';
+import type { Decorator } from '@storybook/react';
```
ArrayKnight, line 17: Please alpha order imports

packages/design-system/.storybook/preview.tsx
```diff
@@ -0,0 +1,37 @@
...
+        color: /(background|color)$/i,
+        date: /Date$/i,
+      },
+    },
+    backgrounds: {
+      default: 'dark',
+      values: [{ name: 'dark', value: 'rgb(11,11,11)' }],
+    },
+    layout: 'centered',
+    actions: { argTypesRegex: '^on.*' },
```
ArrayKnight, line 32: There's going to be a potential issue with automatically binding to all events. There's one (need to find the comment in our code) in RAC that changes the behavior of the component

packages/design-system/package.json
```diff
@@ -39,8 +39,8 @@
     "docs": "pnpm typedoc --logLevel Error --entryPoints src/index.ts --out documentation/api --readme none --disableSources --plugin typedoc-plugin-markdown",
     "index": "pnpm zx ../../scripts/indexer.mjs packages/design-system --barrels",
     "lint": "pnpm biome lint ./",
-    "preview": "pnpm ladle serve",
-    "test": "pnpm vitest --dir=src"
+    "test": "pnpm vitest --dir=src",
+    "storybook": "storybook dev -p 6006"
```
ArrayKnight, line 43: I would prefer to keep the script as `preview`

packages/design-system/package.json
```diff
@@ -63,9 +63,15 @@
     "@accelint/biome-config": "workspace:*",
     "@accelint/typescript-config": "workspace:*",
     "@accelint/vitest-config": "workspace:*",
+    "@chromatic-com/storybook": "^3.2.3",
```
ArrayKnight, line 66: Are we actually integrating with this?

packages/design-system/package.json
```diff
@@ -63,9 +63,15 @@
...
     "@fontsource/roboto-flex": "5.1.0",
     "@fontsource/roboto-mono": "5.1.0",
-    "@ladle/react": "4.1.2",
+    "@storybook/addon-actions": "7.6.20",
```
ArrayKnight, line 69: Actions is already part of essentials

packages/design-system/src/components/button/button.stories.tsx
```diff
@@ -73,174 +36,71 @@ export const ButtonExample: Story<ButtonProps> = ({
...
+        </Icon>
+      ) : (
+        props.children
+      )}
+    </LinkButton>
+  ),
+};
 
-UncontrolledToggleExample.storyName = 'Toggle / Uncontrolled';
+export const Toggle: StoryObj<ToggleButtonProps> = {
```
ArrayKnight, line 87: Why did you get rid of the uncontrolled / controlled examples?

ArrayKnight (changes requested):

packages/design-system/.storybook/preview.tsx
orteth01: this is a remnant of some experimentation I was doing and isn't actually necessary. we're explicitly specifying actions to capture in the stories. i will remove this.

packages/design-system/.storybook/components.tsx
orteth01: Screenshot 2025-01-15 at 8 30 44 AM

ts complains without it - seems like a config issue. was planning on looking into it this am

packages/design-system/.storybook/components.tsx
orteth01: it would be nice if the formatter did this for us automatically 👀

packages/design-system/package.json
orteth01: ts is unhappy if we try to import from this package without specifying it explicitly as a dep.
Screenshot 2025-01-15 at 8 35 15 AM

packages/design-system/package.json
orteth01: not currently. maybe one day? will remove for now though

packages/design-system/src/components/button/button.stories.tsx
orteth01: individual stories didn't feel like the right place to differentiate between uncontrolled / controlled versions of components.

i was thinking we could mimic react-spectrum, and have general docs on the topic instead e.g. https://react-spectrum.adobe.com/react-spectrum/forms.html#controlled-forms

wdyt?

packages/design-system/package.json
orteth01: updated: https://github.com/gohypergiant/standard-toolkit/pull/167/commits/c8c00012ed9518463c70270bbedc2fdd8c37f677

alternatively, wdyt about calling it `dev`?

packages/design-system/package.json
orteth01: removed: https://github.com/gohypergiant/standard-toolkit/pull/167/commits/5a076e8a950d0f4a9e7de269a5a37b6fd50be3e7

packages/design-system/.storybook/preview.tsx
orteth01: cleaned up here: https://github.com/gohypergiant/standard-toolkit/pull/167/commits/0789ff6d84c8a8f1508b6b01bc9efd609fa4224d

packages/design-system/.storybook/components.tsx
orteth01: updated here: https://github.com/gohypergiant/standard-toolkit/pull/167/commits/47318c45b5c96c3f8017dcd3f5a87ad99499423c

packages/design-system/.storybook/components.tsx
ArrayKnight: Absolutely agree

orteth01: looks like biome has a

packages/design-system/src/components/button/button.stories.tsx
switzerb: I disagree - the stories function as a way to demonstrate usage (among other things) and I think it's useful to devs to be able to see both side by side in the ecosystem. I am not opposed to having more general guidance in a central location but I don't think removing the explicit story is a good idea. I am open to hearing the downside of keeping them, but so far the feedback is to have more and more comprehensive stories, not fewer.

orteth01: okay, i'm happy to add controlled stories back!

orteth01: added here: https://github.com/gohypergiant/standard-toolkit/pull/167/commits/0506a2c5a869c8385a12a6d8bac3c63e11e100d0

switzerb:
Closing this ticket because it got lapped by the move to the new design system.

PR #171, opened 2025-01-16
feat: non-react playgrounds for docs

orteth01:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

- add a `@playground` block JSDoc comment in package code
- `pnpm --filter=@accelint/docs run dev
- verify that the playground is added to the document you expect

see `packages/converters/src/boolean-to-number/index.ts` for an example of a `@playground` comment

- [ ] Yes
- [x] No

## 💬 Other information

- using [sandpack](https://sandpack.codesandbox.io/docs) for the playground component (ty @kalisjoshua for the find)
- i'm open to any ideas or feedback on the formatting of the `@playground` comment and how it works.

### DIFF
--- a/apps/docs/lib/playground-plugin-hypergiant.mjs
+++ b/apps/docs/lib/playground-plugin-hypergiant.mjs
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * TypeDoc plugin for replacing @playground blocks with our custom playground component,
+ *
+ * @param {import('typedoc').Application} app
+ */
+export function load(app) {
+  app.converter.on('createSignature', convertBlock);
+}
+
+function convertBlock(context, reflection) {
+  const playgroundTag = reflection?.comment?.blockTags?.find(
+    ({ tag }) => tag === '@playground',
+  );
+
+  if (playgroundTag) {
+    const code = playgroundTag.content[0].text;
+    playgroundTag.content = [
+      {
+        kind: 'text',
+        text: `import { Playground } from '@site/src/components/playground';
+
+<Playground
+  code={\`${code.trim()}\`}
+  dependencies={['${context.project.name}']}
+/>
+        `,
+      },
+    ];
+  }
+}
--- a/apps/docs/src/components/playground.tsx
+++ b/apps/docs/src/components/playground.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Hypergiant Galactic Systems Inc. All rights reserved.
+ * Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
  * This file is licensed to you under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License. You may obtain a copy
  * of the License at https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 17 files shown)


packages/constants/typedoc.mjs
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/** @type {Partial<import("typedoc").TypeDocOptions>} */
+export default {
+  name: '@accelint/constants',
```
orteth01, line 15: i think there may be a bug in typedoc related to using the 'packages' entry point strategy. for whatever reason, context.project.name in the `createSignature` event was an empty string, unless we explicitly set the name like this in each package. in theory, it should just be getting this from package.json....

apps/docs/typedoc.mjs
```diff
@@ -27,7 +29,12 @@ export default {
   pageTitleTemplates: {
     member: (args) => args.name, // simpler page titles for member pages
   },
-  plugin: ['typedoc-plugin-markdown', './lib/typedoc-plugin-hypergiant.mjs'],
+  plugin: [
+    'typedoc-plugin-markdown',
+    './lib/typedoc-plugin-hypergiant.mjs',
+    './lib/playground-plugin-hypergiant.mjs',
```
orteth01, line 35: i kind of like the separation of concerns, but i'm open to including this in the general typedoc-plugin-hypergiant if that's what folks prefer

apps/docs/src/components/playground.tsx
```diff
@@ -0,0 +1,35 @@
...
+  dependencies?: string[];
+};
+
+export function Playground({ code, dependencies = [] }: PlaygroundProps) {
+  return (
+    <Sandpack
+      template='vanilla-ts'
+      customSetup={{
+        dependencies: Object.fromEntries(
+          dependencies.map((dep) => [dep, 'latest']),
```
orteth01, line 26: i just pinned to latest... i'm open to a more complex/flexible structure for the `dependencies` prop, if that's preferable

apps/docs/lib/playground-plugin-hypergiant.mjs
```diff
@@ -0,0 +1,42 @@
...
+        text: `import { Playground } from '@site/src/components/playground';
+
+<Playground
+  code={\`${code}\`}
+  dependencies={['${context.project.name}']}
+/>
+        `,
+      },
+    ];
+  }
```
orteth01, line 41: ended up being fairly simple
- look for the `@playground` tag
- if it's there, replace the content w/ our playground component, passing the original content of the block as the `code` prop

packages/converters/src/boolean-to-number/index.ts
```diff
@@ -16,11 +16,13 @@
...
- * booleanToNumber(true);
+ * @playground
+ * import { booleanToNumber } from '@accelint/converters';
+ *
+ * console.log(booleanToNumber(true));
  * // 1
  *
- * booleanToNumber(false);
+ * console.log(booleanToNumber(false));
  * // 0
```
orteth01, line 26: example `@playground` block

packages/math/typedoc.mjs
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/** @type {Partial<import("typedoc").TypeDocOptions>} */
+export default {
+  name: '@accelint/constants',
```
orteth01, line 15:
```suggestion
  name: '@accelint/math',
```

apps/docs/typedoc.mjs
ArrayKnight: Is there a reason you're adding the `-hypergiant` suffix?

orteth01: not really - just copying prior art. happy to change it if you prefer

kalisjoshua: For this I would suggest `typedoc-plugin-playground.mjs` instead. I made the first one with the thought of it being filled with very small - possibly one-liners - of customization that would only really make sense for us. This one might be something we could generalize to be valuable to the community.

packages/constants/typedoc.mjs
kalisjoshua: Interesting. Not too surprising.

packages/converters/src/boolean-to-number/index.ts
kalisjoshua: This is awesome.

apps/docs/typedoc.mjs
orteth01: updated here: https://github.com/gohypergiant/standard-toolkit/pull/171/commits/83b38b33cd259c3fc4d02daf007ea9fb02239a53

PR #179, opened 2025-02-05
Feat: Remove requirement for listsAsArrays in Querybuilder

switzerb:
Closes 

There is a boolean prop available `listsAsArrays` on the [React Querybuilder](https://react-querybuilder.js.org/docs/components/querybuilder#listsasarrays) component that changes the structure of the data store for those operators that store multiple values. When enabled, `listsAsArrays` stores the data as an array of values and when disabled, the list is stored as a comma-separated string.

Previously, the `listsAsArrays` value was locked into `true` to simplify some of the processing and validation, but since there is a desire to have more flexibility with the filtering (for example, there are CQL queries where the value can be passed through as comma separated strings without any further processing) then this PR makes `listsAsArrays` a prop that can be set per instance.

Value of a multi-value rule with `listsAsArrays` enabled: 
CleanShot 2025-02-05 at 11 30 53

Value of a multi-value rules with `listsAsArrays` disabled:
CleanShot 2025-02-05 at 11 32 33

To be able to handle this isomorphically, RQB provides a hook `useValueEditor` that normalizes the value into an array for processing (no matter how it is stored) which we use to draw the value editor components:

https://github.com/gohypergiant/standard-toolkit/blob/ed5ca137073b9f93eb0eb4c74fa043d9ba8529ac/packages/design-system/src/components/query-builder/value-editor.tsx#L203

Note that whether `listsAsArrays` is set to true or false, the value is always processed as an array in the `ValueEditor` component above:

Value with `listsAsArrays` enabled:
CleanShot 2025-02-05 at 11 40 27

Value with `listsAsArrays` disabled:
CleanShot 2025-02-05 at 11 38 00

The only real downstream effect of this change is on validation, which needs differentiated handling on the part of the application developer.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

To see this change in action, run the design system stories off this branch and navigate to the querybuilder story. There is now a control that allows for toggling the `listsAsArrays` on and off in the story. Note that the validation only works in the story when the `listsAsArrays` is enabled.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/design-system/src/components/query-builder/query-builder.stories.tsx
+++ b/packages/design-system/src/components/query-builder/query-builder.stories.tsx
@@ -81,8 +81,8 @@ export default {
       control: {
         type: 'boolean',
       },
-      defaultValue: true
-    }
+      defaultValue: true,
+    },
   },
 } satisfies StoryDefault<QueryBuilderProps>;
 
--- a/packages/design-system/src/components/query-builder/query-builder.stories.tsx
+++ b/packages/design-system/src/components/query-builder/query-builder.stories.tsx
@@ -77,6 +77,12 @@ export default {
       options: ['sm', 'lg'],
       defaultValue: 'sm',
     },
+    listsAsArrays: {
+      control: {
+        type: 'boolean',
+      },
+      defaultValue: true
+    }
   },
 } satisfies StoryDefault<QueryBuilderProps>;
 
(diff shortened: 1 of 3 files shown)


switzerb:
I'm closing this PR. For the time being, this is not the path that we are going down. We can revive the idea later if needed.

PR #183, opened 2025-02-21
Feat: Add read-only tree node property

switzerb:
Closes https://github.com/gohypergiant/standard-toolkit/issues/186

Adds the concept of a read-only tree node. Note that the story (and the visual choices to indicate "read only") are just used here to demonstrate the concept and can be themed per the application.

CleanShot 2025-02-26 at 09 49 22@2x

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run a preview of the design system stories and take the new property for a whirl. Note that if a read-only child has a parent that is NOT read only, then it is still possible to change the visibility of that item via the parent. For a truly "this can never be touched" the item either needs to be standalone or in a group that is also read-only.

- [ ] Yes
- [x] No


belsrc:
❤️

belsrc (approved):

PR #240, opened 2025-03-20
fix: select popover states

### LINKED ISSUE #239: Popover (Options) does not position Optionslist properly given no space in view.
### Provide a general summary of the issue here

Popover does not work correctly with Options List.

### 🤔 Expected Behavior?

Select drop down should recognize lack of space and position dropdown on top if on bottom of page.

### 😢 Current Behavior

Image

### 🤩 Possible Solution

display: content is causing the Popover positioning to lose reliance on ListBox element for positioning.

```
  list: {
    container: style({
      '@layer': {
        [layers.components.l1]: {
          containerName: optionsContainers.list,
          display: 'content',
        },
      },
    }),
```

### Participation

- [x] I am willing to submit a pull request for this change.

### 📜 Context

_No response_

### 🖥️ Steps to Reproduce

To reproduce:

1. Go select stories and create a div 750px on top of page so select is on bottom of page
2. Click on select
3. Dropdown does not position itself on top of select though it should be the default behavior 

### Which package(s) are the source of the bug?

design-system

### Version

latest

### What browsers are you seeing the problem on?

Chrome

### If other, please specify.

_No response_

### What operating system are you using?

Mac

zhhjoseph:
Closes #239 

## ✅ Pull Request Checklist:
- [x] Included link to corresponding [GitHub Issue](https://github.com/gohypergiant/standard-toolkit/issues/239).

## 📝 Test Instructions:

1. Add a large div on top of any story in packages/design-system/src/components/select/select.stories.tsx.
2. run `pnpm --filter=@accelint/design-system preview`
3. change window so select input will be on bottom of page
4. select dropdown should appropriately position itself on top if there is no space on bottom

- [ ] Yes
- [x] No

### DIFF
--- a/packages/design-system/src/components/options/options.css.ts
+++ b/packages/design-system/src/components/options/options.css.ts
@@ -141,7 +141,7 @@ export const optionsClassNames: OptionsClassNames = {
       '@layer': {
         [layers.components.l1]: {
           containerName: optionsContainers.list,
-          display: 'contents',
+          display: 'block',
         },
       },
     }),
--- a/.changeset/eight-suns-type.md
+++ b/.changeset/eight-suns-type.md
@@ -1,0 +1,5 @@
+---
+"@accelint/design-system": patch
+---
+
+OptionList enable popover behavior
--- a/packages/design-system/src/components/menu/menu.css.ts
+++ b/packages/design-system/src/components/menu/menu.css.ts
@@ -136,7 +136,7 @@ export const menuClassNames: MenuClassNames = {
       '@layer': {
         [layers.components.l1]: {
           containerName: menuContainers.list,
-          display: 'contents',
+          display: 'block',
         },
       },
     }),


packages/design-system/src/components/options/options.css.ts
```diff
@@ -141,7 +141,7 @@ export const optionsClassNames: OptionsClassNames = {
       '@layer': {
         [layers.components.l1]: {
           containerName: optionsContainers.list,
-          display: 'contents',
+          display: 'block',
```
zhhjoseph, line 144: Thought about enabling to override this property but since this should be always be the default behavior thought content should just be replaced.

switzerb: This is such a great find! Will you make the same change to the `menu.css.ts` file in the same place? This is probably the underlying issue with the menu as well.

PR #264, opened 2025-04-23
feat: initial scaffold for @accelint/design-toolkit

switzerb:
Closes https://github.com/gohypergiant/standard-toolkit/issues/269

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Some ways to validate the scaffolding. Ideally, you will do all three:

#### Running locally

To run the local Storybook instance

```bash
pnpm i

pnpm --filter=@accelint/design-toolkit preview
```

If you run into errors, try running
`pnpm build` after you install the dependencies and then run the preview command again. If you don't have pnpm enabled then you can enable it with the command
[`corepack enable pnpm`](https://pnpm.io/installation#using-corepack) .

#### Run a local build

```bash
pnpm turbo --filter=@accelint/design-toolkit build
```

#### Run local example app

To demonstrate usage in a Nextjs app:

```bash
pnpm --filter "*apps/next" run dev
```

- [ ] Yes
- [x] No


design-toolkit/components/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "**/*.bench.ts",
+    "**/*.test.ts",
+    "**/*.test-d.ts"
+  ],
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "rootDir": "src",
+    "outDir": "dist",
+    "paths": {
+      "@/*": ["./src/*"]
```
switzerb, line 16: I was unable to get aliasing to work with Storybook

design-toolkit/components/.storybook/main.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * This function is used to resolve the absolute path of a package.
+ * It is needed in projects that use Yarn PnP or are set up within a monorepo.
+ */
+function getAbsolutePath(packageName: string): string {
+  return dirname(require.resolve(join(packageName, 'package.json')));
+}
+
+const config: StorybookConfig = {
```
switzerb, line 24: For some reason, aliasing imports is not being registered correctly with Storybook -- something to investigate.

design-toolkit/components/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "@types/react": "^19.0.10",
+    "@types/react-dom": "^19.0.4",
+    "autoprefixer": "10.4.20",
+    "esbuild-plugin-file-path-extensions": "2.1.3",
+    "jsdom": "25.0.1",
+    "postcss": "8.4.47",
+    "storybook": "^8.6.12",
+    "tsup": "8.3.0",
+    "typescript": "~5.7.2",
+    "vite": "5.4.9",
```
switzerb, line 74: These are not the latest versions of vite and vitest -- they are this version to be in line with the rest of the monorepo and the centralized tooling configuration. Do we think that it is worth upgrading the packages?

design-toolkit/components/package.json
```diff
@@ -0,0 +1,148 @@
...
+    "build": "pnpm index && pnpm tsup && pnpm build:css",
+    "build:css": "npx @tailwindcss/cli -i ./src/index.css -o ./dist/styles.css",
+    "dev": "pnpm tsc --watch",
+    "index": "pnpm zx ../../scripts/indexer.mjs design-toolkit/components",
+    "lint": "pnpm biome lint",
+    "preview": "storybook dev -p 6006",
+    "test": "pnpm vitest --dir=src",
+    "build-storybook": "storybook build"
+  },
+  "dependencies": {
```
brandonjpierce, line 102: We are going to want to establish our peer deps as well

design-toolkit/components/package.json
```diff
@@ -0,0 +1,148 @@
...
+    "./styles": {
+      "import": {
+        "default": "./dist/index.css"
+      },
+      "default": "./dist/index.css"
+    }
+  },
+  "scripts": {
+    "bench": "pnpm vitest bench --run --dir src",
+    "build": "pnpm index && pnpm tsup && pnpm build:css",
```
brandonjpierce, line 93: `index` script is already a prerequisite in the turbo config, so this will make that run twice as is. We should probably update our docs across the board but you can do something like:

```
turbo build --filter=@accelint/design-toolkit
```

Which would make `index` run before `build`

design-toolkit/components/src/components/avatar/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  className,
+  children,
+  fallback,
+  fallbackDelay,
+  source,
+  size,
+  ...props
+}: AvatarProps) => (
+  <div
+    className={cn([
```
brandonjpierce, line 53: If a `cn` call is completely static e.g. no dynamic values, can it be hoisted outside of the react component?

design-toolkit/components/src/index.ts
```diff
@@ -0,0 +1,113 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * THIS IS A GENERATED FILE. DO NOT ALTER DIRECTLY.
+ */
+
+export { Avatar } from './components/avatar';
```
brandonjpierce, line 17: Were these intended to be moved to the `design-toolkit/icons/` dir instead?

design-toolkit/components/vite.config.ts
```diff
@@ -0,0 +1,26 @@
+/*
```
brandonjpierce, line 1: What is the vite config needed for?

design-toolkit/components/package.json
switzerb: Giving this a go caused a recursive build and then I found this [on the turbo docs site](https://turborepo.com/docs/crafting-your-repository/running-tasks#using-scripts-in-packagejson): CleanShot 2025-04-28 at 11 10 11

I removed the index run and did the build at the root of the site and the indexer ran. But not when running just this command `pnpm run build --filter@accelint/design-toolkit`. Which honestly makes sense to me since we will mostly be building from the root and the index can be run manually.

design-toolkit/components/package.json
switzerb: Good point. Because of the previous confusion about importing sometimes directly from `react-aria-components` and sometimes from the `@accelint/design-system`, the goal this time was to encapsulate more thoroughly -- the expectation being that no application would import `react-aria-components` directly. Or, if it did, there would be no expectation the `context` of the two libraries would work together.

In other words, given the above, maybe having the `react-aria-components` as a `peer` isn't necessary anymore?

design-toolkit/components/src/index.ts
switzerb: Yes! Since this PR already has a lot of moving parts, I thought it would be smarter and simpler to tackle the icons in a separate fast-follow PR, replace/move these icons as a part of that. https://gohypergiant.atlassian.net/browse/PATH-358

design-toolkit/components/package.json
brandonjpierce: When `tsup` goes to bundle things it will automatically mark anything in `dependencies` or `peerDependencies` as external, meaning it does not bundle that dependency directly into our `dist` output. This means that consumers of the library still need to include them as a dependency for things to work. 

Generally speaking you almost never want to bundle in externals directly thus requiring the need to mark the peer deps.

design-toolkit/components/vite.config.ts
switzerb: For storybook, since it extends the `vite.config` file. This allows the @ alias and the tailwind styles to work. `design-toolkit/components/.storybook/main.ts`

switzerb: CleanShot 2025-04-28 at 11 36 32

brandonjpierce: Ahhhh I see I see

design-toolkit/components/package.json
belsrc: > But not when running just this command pnpm run build --filter@accelint/design-toolkit.

You are bypassing Turbo with that command which means it doesn't do the downstream commands. You can use Turbo's version for that, what Brandon linked above.  https://turborepo.com/docs/reference/run#--filter-string

design-toolkit/components/package.json
```diff
@@ -0,0 +1,153 @@
...
+      "import": {
+        "default": "./dist/index.css"
+      },
+      "default": "./dist/index.css"
+    }
+  },
+  "scripts": {
+    "bench": "pnpm vitest bench --run --dir src",
+    "build": "pnpm tsup && pnpm build:css",
+    "build:css": "npx @tailwindcss/cli -i ./src/index.css -o ./dist/styles.css",
```
brandonjpierce, line 94: Small nit: `pnpm dlx` is the equivalent command for `npx`. Alternatively we should maybe just define `@tailwindcss/cli` as a devDependency and call it like so `pnpm @tailwind/cli ...` so that we are not downloading anything during pipeline runs

design-toolkit/components/src/components/avatar/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+'use client';
```
brandonjpierce, line 13: `'use client'` directive will be stripped during bundling IIRC, instead we should probably just `import 'client-only';` for this package.

design-toolkit/components/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "@radix-ui/react-avatar": "^1.1.4",
+    "@react-types/shared": "3.25.0",
+    "react": "^18.0.0 || ^19.0.0",
+    "react-aria-components": "1.7.1",
+    "react-dom": "^18.0.0 || ^19.0.0",
+    "tailwindcss": "^4.1.3"
+  },
+  "dependencies": {
+    "clsx": "^2.1.1",
+    "cva": "npm:class-variance-authority@^0.7.1",
```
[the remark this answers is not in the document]
switzerb: I think it is just for aliasing, temporarily -- from the [docs](https://cva.style/docs/getting-started/installation) while he waits for his cool new npm package name.  I was not aware you could alias packages this way, honestly. It looks like it won't be needed for the `1.0.0` release but it is still in beta.

CleanShot 2025-04-29 at 08 31 09

design-toolkit/components/src/components/avatar/index.tsx
switzerb: whoop -- it makes storybook cranky. taking a look.

CleanShot 2025-04-29 at 08 36 47

design-toolkit/components/package.json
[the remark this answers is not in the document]
shadyendless: This is used so that we can alias it as `cva`:
```ts
import { cva } from 'cva';
```

Rather than:
```ts
import { cva } from 'class-variance-authority';
```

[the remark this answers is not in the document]
brandonjpierce: TIL!

design-toolkit/components/src/components/avatar/index.tsx
brandonjpierce: > Whoa -- I was not familiar with this. Does this have the same effect as `use client` but won't get stripped?

It uses a package.exports trick to throw an exception at runtime basically, so has a similar-ish effect as `use client` but acts more as a developer hint. This means that in your application land you _would_ still need `'use client'`;

switzerb: I totally thought 'client-only' was one of nextjs's magic strings.

.github/workflows/gh-pages.yml
```diff
@@ -32,9 +32,9 @@ jobs:
           cache: "pnpm"
       - run: pnpm install
       - run: pnpm run build
-      - name: Ladle
+      - name: Ladle/Storybook
         id: build
-        run: pnpm --filter=@accelint/design-system exec ladle build
+        run: pnpm build:design-preview
```
gspnz, line 37: Does this change just add building the storybook output but continues to load the ladle instance as the source artifact?

If so, will the shift to public storybook come in a future PR?

apps/next/package.json
```diff
@@ -13,29 +13,32 @@
...
     "jsdom": "25.0.1",
-    "react-test-renderer": "18.3.1",
     "typescript": "5.6.3",
     "vite": "5.4.9",
     "vitest": "2.1.3"
   },
   "dependencies": {
     "@accelint/bus": "workspace:*",
     "@accelint/design-system": "workspace:*",
+    "@accelint/design-toolkit": "workspace:*",
```
kalisjoshua, line 32: Is the duplication of "design-system" and "design-toolkit" something that will go away after our transition period or is it an forever thing I just don't understand?

design-toolkit/components/src/components/box/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  direction?: 'row' | 'col';
+  /** Horizontal alignment in flex container */
+  justify?: 'start' | 'center' | 'end' | 'between' | 'around' | 'evenly';
+  /** Vertical alignment in flex container */
+  align?: 'start' | 'center' | 'end' | 'stretch' | 'baseline';
+  /** Gap between children (matches Tailwind's spacing scale) */
+  gap?: number;
+}
+
+export const Box = ({
```
kalisjoshua, line 29: Just curious why we rename `div` to `Box` and not `Div`? This specifically and in general why do we rename things that people would already know?

design-toolkit/components/src/components/classification-badge/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  },
+);
+
+export interface ClassificationBadgeProps
+  extends VariantProps<typeof classificationBadge> {
+  className?: string;
+  children?: ReactNode;
+}
+
+export const ClassificationBadge = ({
```
kalisjoshua, line 54: For classification badge/banner we could just build all the explicit components because they are limited in number and unlikely to change in number or content. It is a small optimization but small things add up and this is a low risk component to optimize.

Or am I missing something?

design-toolkit/components/src/index.ts
```diff
@@ -0,0 +1,103 @@
+/**
+ * THIS IS A GENERATED FILE. DO NOT ALTER DIRECTLY.
```
kalisjoshua, line 2: Isn't this a barrel file? Weren't we going to move away from barrel files?

kalisjoshua:
Personally not a fan of `.keep` files but overall I think it looks good. I did not run it locally and check deeply so if that is necessary let me know and I will.

apps/next/package.json
switzerb: Yeah, just a temporary thing. I wanted to science whether the two could co-exist in the same nextjs site so that when the time comes to migrate, it doesn't have to be a zero-sum game. As long as they can run in the same repo for a little while, the migration and deprecation of the design-system should be much simpler

design-toolkit/components/src/index.ts
brandonjpierce: Primarily in application land (next, vite, parcel), for package land it is fine so long as we can guarantee proper tree shaking is still occurring

.github/workflows/gh-pages.yml
shadyendless: It does both, builds ladle, builds Storybook, and moves Storybook to be a subdirectory of the Ladle deployment so we have access to both:
`pnpm --filter=@accelint/design-system exec ladle build && pnpm --filter=@accelint/design-toolkit build-storybook -o ../../packages/design-system/build/design-toolkit`

gspnz: For sure, the output step in this workflow though references the ladle artifact, right? It looks like it remains unchanged

design-toolkit/components/src/components/box/index.tsx
shadyendless: `Box` is a very common layout utility across a lot of UI frameworks that people are already familiar with: 
- https://mui.com/material-ui/react-box/
- https://chakra-ui.com/docs/components/box
- https://mantine.dev/core/box/
- https://v2.grommet.io/box

Figured we should keep with that pattern.

design-toolkit/components/src/components/classification-badge/index.tsx
shadyendless: Not sure what you're mentioning here, unless you mean breaking it out into something like this?
```ts
<MissingClassificationBadge />
<CuiClassificationBadge />
<TSSCIClassificationBadge />
```

design-toolkit/components/src/index.ts
shadyendless: It's used in conjunction with targeted exports to let consumers opt into the version they want to use, either:
```ts
import { Button } from '@accelint/design-toolkit';
import { Button } from '@accelint/design-toolkit/button';
```

shadyendless:
> Personally not a fan of `.keep` files but overall I think it looks good. I did not run it locally and check deeply so if that is necessary let me know and I will.

Good callout, that file will be removed once we move icons over into a separate package (they are currently in the design system but moving out of it).

design-toolkit/components/src/components/box/index.tsx
kalisjoshua: Not sure MUI is a good citation or not, but I understand.

design-toolkit/components/src/components/classification-badge/index.tsx
kalisjoshua: Yes. For these specifically, I'm guessing, there is never going to be a need for a runtime change of any kind so creating them all or only the needed one (or few) at runtime makes sense to me. Not a big deal in any way.

brandonjpierce (approved):

PR #265, opened 2025-04-24
Feature: event bus

brandonjpierce:
Andrew did the bulk of this, I am just bringing it over from another repo and doing slight adjustments.

Follow ons:

- React hook
- AbortController support
- Zustand plugin?

### DIFF
--- a/packages/bus/tsup.config.ts
+++ b/packages/bus/tsup.config.ts
@@ -31,4 +31,5 @@ export default defineConfig({
   sourcemap: true,
   splitting: true,
   treeshake: true,
+  metafile: true,
 });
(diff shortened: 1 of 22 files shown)


packages/bus/tsup.config.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import { esbuildPluginFilePathExtensions } from 'esbuild-plugin-file-path-extensions';
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  esbuildPlugins: [
+    esbuildPluginFilePathExtensions({
+      esmExtension: 'js',
+    }),
```
shadyendless, line 20: On the `design-toolkit` repo, I had to extend this because the aliases weren't being resolved in the built output (so it was including like `import { thing } from @/lib`:

```ts
import { fixImportsPlugin } from 'esbuild-fix-imports-plugin';

{
esbuildPlugins: [
    esbuildPluginFilePathExtensions({
      esmExtension: 'js',
    }),
    fixImportsPlugin(),
  ]
}

shadyendless: Yep, saw the alias in the config so thought I'd call it out.

nrdobie (approved):

PR #272, opened 2025-05-03
Feat/checkbox

shadyendless:
Adds the `<Checkbox>` and `<Checkbox.Group>` component.

https://github.com/user-attachments/assets/f1619ef8-1444-408e-a06c-d836a2c25dcf


design-toolkit/components/src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    >
+      {({
+        isDisabled,
+        isFocused,
+        isHovered,
+        isIndeterminate,
+        isReadOnly,
+        isSelected,
+      }) => {
+        console.log({
```
brandonjpierce, line 99: We good to remove log now?

brandonjpierce:
Nit: we should definitely add a outline when keyboard focused by default with some theme value to allow us to set it to transparent if we want.

brandonjpierce:
Nit: doesn't look like it renders the correct border color in light mode

brandonjpierce:
Confirming tree shaking potential is still looking great

Screenshot 2025-05-07 at 11 00 31 PM

design-toolkit/components/src/components/checkbox/index.tsx
shadyendless: Good catch, removed this.

brandonjpierce (approved):

PR #275, opened 2025-05-05
Feat/add accordion

shadyendless:
Adds the accordion to the design system

https://github.com/user-attachments/assets/903b3510-1c0e-4991-8e7f-05de55e7356e


brandonjpierce:
Nit: focus styles could use some minor adjustment. Also see my note for focus styles on checkbox PR as the style should match here as well (ideally). Also there is no focus for the options button if enabled.

brandonjpierce (approved):

PR #277, opened 2025-05-06
Feat/add chip component

shadyendless:
Adds the `<Chip>`, `<Chip.List>`, `<Chip.Selectable>`, and `<Chip.Deletable>` components.

https://github.com/user-attachments/assets/c11ede33-7049-4eb5-9ef8-aff870dc3454


design-toolkit/components/src/components/chip/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { CancelFill } from '@/icons';
```
brandonjpierce, line 13: Make a mental note to possibly refactor this import later so we can ensure tree shaking is happening post icon migration

brandonjpierce:
@shadyendless for the delete chip, was that suppose to have keyboard interaction for the delete? E.g. pressing space when keyboard focused deletes it?

brandonjpierce:
@shadyendless minor nit: the font size between each of the variations is slightly different, this may be reflected in design but we should try and persuade for a common look and feel.

shadyendless:
> @shadyendless for the delete chip, was that suppose to have keyboard interaction for the delete? E.g. pressing space when keyboard focused deletes it?

It's a bit confusing, but backspace is how you delete the chips that are focused. I didn't modify the base behavior of RAC on that front.

shadyendless:
> @shadyendless minor nit: the font size between each of the variations is slightly different, this may be reflected in design but we should try and persuade for a common look and feel.

Sounds like an oversight on my behalf, good catch.

design-toolkit/components/src/components/chip/index.tsx
shadyendless: Yep, have a follow-up once the icons are in place to remove them from this repository, should see that come through early next week.

brandonjpierce (approved):

brandonjpierce (approved):

PR #278, opened 2025-05-07
Feat/add inputs

shadyendless:
Adds the `<TextField>` component.

https://github.com/user-attachments/assets/235209d8-cd95-4048-b292-03b693c5c972


.changeset/sharp-spiders-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@accelint/design-toolkit': patch
+---
+
+Added the initial scaffolding for documentation to the existingcomponents.
```
brandonjpierce, line 5:
```suggestion
Added the initial scaffolding for documentation to the existing components.
```

design-toolkit/components/src/components/text-field/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  type InputProps as AriaInputProps,
+  Text as AriaText,
+  TextField as AriaTextField,
+  type TextFieldProps as AriaTextFieldProps,
+  Button,
+  InputContext,
+  useContextProps,
+} from 'react-aria-components';
+
+import { CancelFill } from '@/icons';
```
switzerb, line 25: depending on what gets merged first, this is going to have to change to the `@accelint/icons` package

design-toolkit/components/src/components/text-field/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  isInvalid,
+  isReadOnly,
+  label,
+  placeholder,
+  size = 'medium',
+  ...props
+}: TextFieldProps) {
+  const isSmall = size === 'small';
+  const shouldShowDescription = !(isSmall || isInvalid) || isDisabled;
+  const shouldShowError = isInvalid && !isDisabled && !isReadOnly;
```
switzerb, line 201: might want to add a check here for whether `errorMessage` is defined/has content. same with the description, I suppose

design-toolkit/components/src/components/text-field/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+import { type VariantProps, cva } from 'cva';
+import { type ChangeEvent, type ForwardedRef, useEffect } from 'react';
+import { Label } from '../label';
+
+const textFieldStyles = cva(
+  [
+    'block w-full rounded-medium px-s py-xs font-display outline outline-interactive',
+  ],
+  {
+    variants: {
```
switzerb, line 36: not sure this is clever, but one of the variants here could come from the prop `isFocusWithin` that would allow it to have focused props if it's composed parent has focus...something like a dialog with a textfield inside it.

```
    isFocusWithin: {
      false: 'border-stuff-nope',
      true: 'border-stuff-yep',
    },```

File: design-toolkit/components/src/components/text-field/index.tsx [hunk 5]
```diff
@@ -0,0 +1,247 @@
...
+            isClearable: false,
+            isDisabled: false,
+            isReadOnly: props.readOnly,
+            size,
+            className,
+          }),
+        )}
+      >
+        {props.value}
+      </span>
```
switzerb, line 130: I'm wondering if it is worth having an empty string when the field is read-only without any content, to make it more obvious that there is no value? it's sort of a weird edge case.

CleanShot 2025-05-15 at 12 47 44

switzerb (commented):
LGTM -- couple of comments. needs a changeset.

design-toolkit/components/src/components/text-field/index.tsx
shadyendless: I added the class `empty:hidden` so that they do not render if there is no message provided. That does add unnecessary markup to the resulting DOM though so I'll include them in this condition.

design-toolkit/components/src/components/text-field/index.tsx
shadyendless: I don't know if it applies to this component, but it is definitely useful on other things (I think the `accordion` uses it for styling).

design-toolkit/components/src/components/text-field/index.tsx
shadyendless: Good catch, I had that in there and then removed it for some reason.

switzerb (dismissed):

switzerb (approved):

PR #280, opened 2025-05-07
Feat/add switch to design tk

shadyendless:
Adds the `<Switch>` component to the design toolkit.

https://github.com/user-attachments/assets/217a7d12-5d48-4db1-9d66-a9490ba68297


design-toolkit/components/src/components/switch/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <AriaSwitch
+      {...props}
+      className='group flex ai-disabled:cursor-not-allowed items-center gap-s'
+    >
+      {({ isDisabled, isSelected }) => (
+        <>
+          <div
+            className={cn(switchStyles({ className, isDisabled, isSelected }))}
+          />
+          <span className={Label.as()}>{children}</span>
```
brandonjpierce, line 66: Minor question as I have not tested it:

Since this is a static output technically, can the function call be hoisted out of the react component e.g.

```
const labelClasses = Label.as();

... in react component

<span className={labelClasses}>
```

design-toolkit/components/src/index.css
```diff
@@ -343,6 +343,10 @@
   height: fit-content;
 }
 
+@utility w-* {
+  width: --value([ *]);
```
brandonjpierce, line 347: This is to allow for things like `w-[32px]` correct? If so tailwind does ship this by default: https://tailwindcss.com/docs/width#using-a-custom-value

shadyendless: This is but we reset the spacing utility as part of the theme, and so this is to "recreate" that functionality.

brandonjpierce (approved):

PR #281, opened 2025-05-08
feat: add the textarea component.

shadyendless:
Added the `<TextArea>` component to the Design Toolkit.

https://github.com/user-attachments/assets/a1729fae-1cb7-4353-8b23-9c69e008229c


switzerb (dismissed):
LGTM

One thing that i notice is that in the "medium" size the label and the text look pretty similar, which is accurate to the designs but looks a little weird to me so might be worth a conversation with Dell. Also we might want to do something with the browser default scrollbar -- looks weird in dark mode.

CleanShot 2025-05-15 at 12 31 02

switzerb (approved):

PR #290, opened 2025-05-23
feat: smeegl cli

belsrc:
## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No


tooling/smeegl/src/lib.ts
```diff
@@ -0,0 +1,39 @@
...
+// But it seems under construction at the moment? Or something isn't exposed properly.
+
+export async function smeegl(
+  rootPath: string,
+  glob: string,
+  out: string,
+  spreet?: string,
+) {
+  if (!glob) {
+    throw new Error('No glob pattern prvided');
```
switzerb, line 29:
```suggestion
    throw new Error('No glob pattern provided');
```

tooling/smeegl/src/main.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  result.match({
+    Ok: () => spinner.succeed(),
+    Err: (e) => spinner.fail(e),
+  });
+
+  return result;
+}
+
+program
```
switzerb, line 120: It might be an interesting optimization to catch this particular error "too many arguments" when the user forgets to turn the blog into a string, like I did at first -- my guess is this is a very common mistake that users will make, even with your example provided.

CleanShot 2025-05-29 at 10 56 32

switzerb (dismissed):
Looks great to me. Couple of small comments but works as advertised and the code is so delightfully readable.

tooling/smeegl/src/main.ts
belsrc: That is in the `commander.js` side of the house. It handles all of the argument parsing. I could wrap the entire app in a `try..catch` but that seems excessive since I would just be outputting the same error message in a different color and doing a `process.exit(1)`

devindavies (approved):

PR #291, opened 2025-05-23
Fix/update text inputs

shadyendless:
- Updates the `<TextField>` and `<TextArea>` components to also accept the props for the underlying `<Input />` components and pass those through. This allows us to receive the native HTML props as well.

- Adds an optional `selectOnFocus` prop that, when `true`, automatically selects the text in the `TextField` or `TextArea` to allow the user to quickly replace the contents.

https://github.com/user-attachments/assets/f5b429d9-0542-4154-87ae-3ce2a1530b02


design-toolkit/components/src/components/text-field/index.tsx
```diff
@@ -227,6 +233,7 @@ export function TextField({
         isClearable={isClearable}
         placeholder={placeholder}
         size={size}
+        {...(props as InputProps)}
```
brandonjpierce, line 236: Was this what was potentially causing the `defaultValue` issue we were seeing on Launchpad? E.g. the state value would update but the `TextField` would not?

shadyendless: Oh, good catch, this very well may have been the issue! I was assuming it would have just worked with context magic but maybe it was getting broken somehow.

PR #293, opened 2025-05-29
feat: add more predicates

belsrc:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

`pnpm run test --filter=@accelint/predicates`

- [ ] Yes
- [x] No

## 💬 Other information

Additional predicates that will be needed for an internal ticket.

### DIFF
--- a/packages/predicates/src/is-equal/index.ts
+++ b/packages/predicates/src/is-equal/index.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Hypergiant Galactic Systems Inc. All rights reserved.
+ * Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
  * This file is licensed to you under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License. You may obtain a copy
  * of the License at https://www.apache.org/licenses/LICENSE-2.0
@@ -10,14 +10,19 @@
  * governing permissions and limitations under the License.
  */
 
-import baseConfig from '@accelint/vitest-config/no-dom';
-import { defineConfig, mergeConfig } from 'vitest/config';
+import { equality } from '@accelint/core';
 
-export default mergeConfig(
-  baseConfig,
-  defineConfig({
-    test: {
-      setupFiles: './src/test.setup.ts',
-    },
-  }),
-);
+/**
+ * Determines if the provided values are strictly equal.
+ *
+ * @param a - The first value to check against.
+ * @param b - The second value to check against.
+ *
+ * @remarks
+ * pure function
+ *
+ * @example
+ * isEqual(5)(32); // false
+ * isEqual('foo')('foo'); // true
+ */
+export const isEqual = equality;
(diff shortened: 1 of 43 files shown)


kalisjoshua (commented):
I will abstain from providing a judgement as I don't agree with the premise but not to the degree that I would prevent it from existing.

belsrc:
In order to construct a compose pipeline, you would have to wrap them anyways. Not much different from doing here or "in app." Which is exactly what we have to do internally to iterate on the layers orchestration. We need to take something like:

```json
  {
    prop: 'icon',
    conditional: {
      combinator: 'AND',
      rules: [{
        field: 'is_deadman',
        operator: '=',
        value: true,
      }],
    },
    result: 9923,
  },
  {
    prop: 'color',
    conditional: {
      combinator: 'AND',
      rules: [{
        field: 'altitude',
        operator: '>=',
        value: 30000,
      }],
    },
    result: [255, 0, 0],
  },

```

and map it to a composition of functions to pass to an accessor.

kalisjoshua:
And this is where I am naive to be honest. I don't know the use case that this is being created for.

However, as far as I know, Javascript is not fully/truly multithreaded (and definitely not by default) so breaking up the operations to this low of a level isn't a profoundly advantageous approach for performance of execution; it may however be better for maintainer readability, which is not to be discredited at all.

I would think that it would be better to abstract functionality at a more meaningful and domain-specific way that could be spread out to a multithreaded utility which would provide a better balance between developer maintainability and performance optimization; even if at the expense of reusability of implementation.

packages/predicates/src/is-equal/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { equality } from '@accelint/core';
+
+/**
+ * Determines if the provided values are strictly equal.
```
switzerb, line 16: Might want to add a note here that this is for integers only.

switzerb:
After learning more from the above comments and our conversation, I am convinced using this model for chaining in our applications will both make the logic much more readable and also make it more resilient. We do a lot of math, use a lot of comparators AND we deal with data streams that can be unpredictable. The property testing makes a lot of sense here -- my only comment is to either be clear about the "integer" constraints or to broaden the testing if these are intended for, say, floats.

packages/predicates/src/is-equal/index.ts
belsrc: Well, can be used with anything. If you are going strictly by the test cases, I can see where that thought would come from. fast-check doesn't have something like "scalar". I suppose that I could doing something like `fc.oneof(fc.int(), fc.float(), fc.string(), fc.boolean())`. Just seems like a lot since it will hold true for any data type given that it is `===`.

shadyendless (approved):

PR #295, opened 2025-06-02
feat/add treeview component to designTK

switzerb:
Closes https://github.com/gohypergiant/standard-toolkit/issues/300.

https://github.com/user-attachments/assets/091781a7-312c-49c1-a0f7-1c3a45d5590d

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

`pnpm install`

`pnpm --filter=@accelint/design-toolkit run preview`

- [ ] Yes
- [ ] No


brandonjpierce:
Design nit: left / right padding is a little off

Screenshot 2025-06-05 at 2 23 23 PM

design-toolkit/components/src/components/tree-view/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { cn } from '@/lib/utils';
+import './tree.css';
```
brandonjpierce, line 14: We may have to reproduce this with tailwind only as it puts an implicit dependency on the consumer to have css module support in place (as well support for 3rd party deps vs. first party).

design-toolkit/components/src/components/tree-view/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+    </button>
+  ) : (
+    <button type='button' onClick={collapse} {...getProps()}>
+      <Icon>
+        <ChevronDown />
+      </Icon>
+    </button>
+  );
+
+  const spacer = (
```
brandonjpierce, line 229: Perf nit: hoist this out of this component since it does not use any dynamic props or state

design-toolkit/components/src/components/tree-view/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+        <Icon>{item.getItemData().iconPrefix}</Icon>
+      )}
+      <div
+        className={cn(
+          'flex-1 items-center text-start',
+          'group-data-[variant=compact]:py-xs group-data-[variant=cozy]:py-s group-data-[variant=tight]:py-xxs',
+        )}
+      >
+        {item.getItemName()}
+        {item.getItemData().description && (
```
brandonjpierce, line 289: Perf nit:

Can set `item.getItemData()` to a var and just re-use it. We may even be able to wrap it in a `useMemo`

```
const obj = useMemo(() => {
  return item.getItemData();
}, [item]);

brandonjpierce:
Design nit: when in tight variant, horizontal spacing is a little inconsistent (specifically on the left hand side

Screenshot 2025-06-05 at 3 02 43 PM

design-toolkit/components/src/components/tree-view/index.tsx
switzerb: I am so, so sad that I agree with you on this observation. Would it be an abomination to simply include it as a part of the `index.css` for tailwind? https://tailwindcss.com/docs/adding-custom-styles#using-custom-css

PR #298, opened 2025-06-03
Feat/add tabs component

sgullen-dev:
Closes [PATH-434](https://gohypergiant.atlassian.net/browse/PATH-434)

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run Storybook
- Check that there is a Tabs folder
- Check that when clicking the Tabs folder, it presents a summary of the Tabs component
- Check that there are sub folders for Tabs, List, and Tab components which demonstrate the props

- [ ] Yes
- [x] No


design-toolkit/components/src/components/tabs/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+export function Tabs({
+  children,
+  className,
+  orientation = 'horizontal',
+  isDisabled = false,
+}: TabsProps) {
+  return (
+    <AriaTabs
+      orientation={orientation}
```
switzerb, line 40:
```suggestion
  isDisabled = false,
  ...rest
}: TabsProps) {
  return (
    <AriaTabs
    ...rest
      orientation={orientation}
```
You'll want to spread the rest of the supported props into the component.

design-toolkit/components/src/components/tabs/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+}
+
+const tabListStyles = cva(
+  'group flex flex-col ai-orientation-horizontal:flex-row',
+  {
+    variants: {
+      isIcons: {
+        true: '[&>*]:leading-[0] [&>*]:p-xs ai-orientation-horizontal:[&>*]:pl-s ai-orientation-horizontal:[&>*]:pr-s',
+        false: '[&>*]:text-header-m [&>*]:p-s',
+        undefined: '[&>*]:text-header-m [&>*]:p-s',
```
switzerb, line 65: wouldn't `undefined` just evaluate to false?

design-toolkit/components/src/components/tabs/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        isDisabled: true,
+        isSelected: true,
+        className: 'fg-disabled bg-interactive-disabled group-ai-orientation-horizontal:border-interactive-disabled',
+      },
+    ],
+  },
+);
+
+export interface TabProps extends AriaTabProps {
+  id: string;
```
switzerb, line 144: `id` is already a prop on `Tab` https://react-spectrum.adobe.com/react-aria/Tabs.html#tab

design-toolkit/components/src/components/tabs/index.tsx
sgullen-dev: That's what I assumed too but it does not

design-toolkit/components/src/components/tabs/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+  label?: string;
+  isIcons?: boolean;
+  isDrawer?: 'left' | 'right' | 'top' | 'bottom';
+}
+
+const tabListStyles = cva(
+  'flex ai-orientation-horizontal:flex-row flex-col',
+  {
+    variants: {
+      isIcons: {
```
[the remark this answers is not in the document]
sgullen-dev: Updated to `variant` and `drawer`, and added type docs to support

brandonjpierce (approved):

design-toolkit/components/src/lib/react.ts
```diff
@@ -27,30 +44,36 @@ export function containsExactChildren({
...
+    if (value < min) { missingComponentsArray.push(`${min - value} of <${key}>`) }
+    if (max !== undefined && value > max) { excessComponentsArray.push(`${value - max} of <${key}>`) }
+  }
 
-  const missingComponents = Object.entries(restrictionResults).filter(
-    ([, value]) => value !== 0,
-  );
+  if (missingComponentsArray.length || excessComponentsArray.length) {
+    const formatList = (label: string, items: string[]) =>
+      items.length ? `\t${label}:\n\t${items.join(', ')}\n` : '';
```
switzerb, line 70: Really like where you are going with this, but I could only get the error to throw by exceeding the max allowed components, while it happily allowed less than the minimum. Am I misunderstanding the intent behind this change?

CleanShot 2025-06-10 at 21 45 24@2x

design-toolkit/components/src/components/tabs/tab-list.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+  component: Tabs.List,
+  args: {
+    variant: 'default',
+    drawer: undefined,
+  },
+  argTypes: {
+    variant: {
+      control: 'select',
+      options: ['default', 'icons'],
+      table: {
```
switzerb, line 29: I am not sure I love the pattern here of locking down the args and providing two stories where they can't be changed. I think I can see the reasoning -- it isn't an "icon" story if you don't pass in the variant "icon" -- but as a user it was confusing and I assumed it was an error until I went and looked at the code. Can we just provide a "With Icons" story and "With Text" story but still let people toggle? I know it might seem counter-intuitive but showing people the args in a control that is *intended* to be interactive then locking it down feels...incorrect somehow.

design-toolkit/components/src/components/tabs/index.tsx
switzerb: I see that you resolved this, but I still am not really sure why you would extend an interface that already has those properties -- unless you intended to lock down the id to a `string` instead of a `Key`? Why? And actually, `isDisabled` is already a boolean property as well. I am just having trouble understanding the reasoning here.

switzerb:
Looks great! I really appreciate the additional docs and love the `mdx` pattern -- definitely going to put more of that into the other components. I am not fully convinced by the multiple stories for `Tabs.List`, `Tabs.Panel` -- etc. I would prefer there to be one `tabs.stories.tsx` myself with just multiple examples. Having them all broken out not only makes me spend time trying to figure out what is different with all the stories, it implies they can all be used as separate pieces, which they can't be. I won't die on this hill, but it feels unnecessarily complex to me for no engineering or developer experience benefit.

Other than confusion with the types, everything else seems solid and is what I would expect. I'll give you the approval since most of my comments are opinions that I would consider non-blocking at this point.

switzerb (approved):

design-toolkit/components/src/lib/react.ts
sgullen-dev: Fixed bug and added test coverage to make sure this function continues working as expected

design-toolkit/components/src/components/tabs/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+  TabPanel as AriaTabPanel,
+  type TabsProps as AriaTabsProps,
+  type TabListProps as AriaTabListProps,
+  type TabProps as AriaTabProps,
+  type TabPanelProps as AriaTabPanelProps,
+} from 'react-aria-components';
+
+export interface TabsProps extends AriaTabsProps {
+  orientation?: 'horizontal' | 'vertical',
+  isDisabled?: boolean,
```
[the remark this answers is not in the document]
sgullen-dev: Good catch, I removed the unnecessary prop definitions when they already exist by extending other types

design-toolkit/components/src/components/tabs/tab-list.stories.tsx
sgullen-dev: Since I am refocusing on the Search component, I think it would be best to circle back around to updating the docs. Gives some time to think it through and decide on a standard for us to follow going forward, too.

switzerb (approved):

PR #304, opened 2025-06-09
feat: add dialog design-tk

thebdewitt:
Closes https://github.com/gohypergiant/standard-toolkit/issues/305

## ✅ Pull Request Checklist:

- [ ] Yes
- [x] No

### DIFF
--- a/design-toolkit/components/src/components/dialog/index.tsx
+++ b/design-toolkit/components/src/components/dialog/index.tsx
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2025 Hypergiant Galactic Systems Inc. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { cn } from '@/lib/utils';
+import { useIsSSR } from '@react-aria/ssr';
+import { cva } from 'cva';
+import {
+  type PropsWithChildren,
+  type ReactNode,
+  type RefObject,
+  createContext,
+  forwardRef,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+import {
+  type DialogRenderProps,
+  type DialogTriggerProps,
+  Dialog as RACDialog,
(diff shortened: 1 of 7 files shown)


design-toolkit/components/src/components/dialog/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      >
+        <RACModal
+          ref={ref}
+          className={cn(
+            'flex flex-col justify-center bg-surface-overlay align-start',
+            size === 'sm' && 'rounded-medium',
+            size === 'lg' && 'rounded-large',
+          )}
+        >
+          <RACDialog className={dialogClasses({ size })}>{children}</RACDialog>
```
switzerb, line 166: You can add `{/* @ts-expect-error package version mismatch TODO */}` above the `children` here to just ignore that `bigint` error you are getting. The TODO is because these ignores will become irrelevant when we remove the older design system.

design-toolkit/components/src/components/dialog/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  className,
+}: PropsWithChildren<{ className?: string }>) => {
+  const { size } = useDialogContext();
+  return (
+    <div
+      className={cn(
+        'flex justify-end gap-xs',
+        //margin is hardcoded to match the spacing values
+        size === 'sm' && 'mt-[16px]',
+        size === 'lg' && 'mt-[24px]',
```
switzerb, line 234: You can actually use tailwind to do these without hardcoding the values, so they are consistent with the theme:

```suggestion
        size === 'sm' && 'mt-l',
        size === 'lg' && 'mt-xl',
```

design-toolkit/components/src/components/dialog/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      onOpenChange,
+      parentRef,
+      isKeyboardDismissDisabled,
+    } = useDialogContext();
+    const isSSR = useIsSSR();
+    const [portal, setPortal] = useState(isSSR ? null : document.body);
+
+    useEffect(() => {
+      const node = parentRef?.current;
+      const port = isSSR ? null : document.createElement('div');
```
switzerb, line 134: This is the detail from the bug that I *think* is caused by the above line of code. I think, stepping through this, that we are going to see the same kind of problems. What is confusing, however, is that I would expect the component to re-render once the hook returns false and we are on the client. I think we might need to build in a test case into the example application, but it might be tricky to know we are solving the right problem, without the help of the dev who uncovered the bug.

I think -- it makes sense to have a separate ticket to manage this. Will you put a TODO comment here so we don't forget?

CleanShot 2025-06-10 at 22 39 35@2x

design-toolkit/components/src/components/dialog/index.tsx
thebdewitt: Good call, I'm not sure how I missed that!

switzerb (approved):

PR #316, opened 2025-06-18
feat: added options component

devindavies:
Closes 

## ✅ Pull Request Checklist:

- [ ] Yes
- [ ] No


devindavies:
https://github.com/user-attachments/assets/fa0fd1f4-c0c7-434b-b04f-0d2e68bb93dc

design-toolkit/components/src/components/combobox/combobox.stories.tsx
```diff
@@ -45,33 +46,33 @@ const meta: Meta<typeof ComboBox> = {
...
+interface CustomOptionsItem extends IOptionsItem {
   isDisabled?: boolean;
 }
 
-const items: CustomMenuItem[] = [
+const items: CustomOptionsItem[] = [
   {
     id: 1,
-    icon: <Placeholder />,
+    leftIcon: <Placeholder />,
```
switzerb, line 56: super nit: in the tree I am calling this a `prefixIcon` since it seems like the most accurate description of what we are doing. Definitely not a blocker for merge but it occurs to me that it would be useful to use the same semantic across all our components in the system.

switzerb:
LGTM. Only thing I have is one naming nit and that the stories have pulled in all the possible props...which is great except they are a little confusing without context or examples. I think it would be better to either lock it down for now (for example, Omit the `dragAndDrop` hooks as a prop until we actually want to support drag and drop in the options list) or at least hide them from the storybook story by only including useful props as control levers: 
CleanShot 2025-06-23 at 10 34 44@2x

PR #326, opened 2025-07-01
Feat/destk table component

devindavies:
Closes [https://gohypergiant.atlassian.net/browse/PATH-440](https://gohypergiant.atlassian.net/browse/PATH-440)

## ✅ Pull Request Checklist:

- [ ] Yes
- [ ] No


design-toolkit/components/src/components/table/table.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+    data: defaultData,
+    columns: columns,
+    showCheckbox: true,
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof Table<Person>>;
+
+export const Default: Story = {
```
switzerb, line 96: This is likely already on your radar, but having robust examples in the Story is essential -- along with as detailed notes as you can stand to write about how/why to use the various probs. I like that you have Tanstack table behind the curtain, but we don't want devs to have to learn the minutia of the underlying headless Table to use this component...at least not in it's simplest incarnation.

design-toolkit/components/src/components/table/table-row.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export type TableRowProps = HTMLAttributes<HTMLTableRowElement> & {
+  ref?: ForwardedRef<HTMLTableRowElement>;
+};
+
+export function TableRow({ ref, className, ...props }: TableRowProps) {
+  return (
+    <tr
+      ref={ref}
+      className={cn(
```
switzerb, line 24: This will need refactoring to use the tailwind-variant library -- which actually I think will simplify some of these tailwind classes

design-toolkit/components/src/components/table/table-body.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import { cn } from '@/lib/utils';
+import type { ForwardedRef, HTMLAttributes } from 'react';
+
+export type TableBodyProps = HTMLAttributes<HTMLTableSectionElement> & {
+  ref?: ForwardedRef<HTMLTableSectionElement>;
+};
+export function TableBody({ className, ref, ...props }: TableBodyProps) {
```
switzerb, line 19: We'll want to continue the pattern of having namespaced components for import: `Table.Body` for example, for these children components that are intended to be composed together.

design-toolkit/components/src/components/table/table-cell.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { cn } from '@/lib/utils';
+import { type VariantProps, cva } from 'cva';
+import type { ForwardedRef, TdHTMLAttributes } from 'react';
+
+export type TableCellProps = TdHTMLAttributes<HTMLTableCellElement> &
+  VariantProps<typeof cellStyles> & {
```
ArrayKnight, line 18: You can use RefAttributes from React instead of having to create this type

design-toolkit/components/src/components/table/table-body.tsx
ArrayKnight: You'd still be able to reference the sub components with out needing to implement the parent Table

design-toolkit/components/src/components/table/index.tsx
```diff
@@ -0,0 +1,666 @@
...
+                                  : moveColumnLeft(header.column.getIndex()),
+                            },
+                            {
+                              label: 'Move column right',
+                              onAction: () =>
+                                header.column.getIsLastColumn('center')
+                                  ? undefined
+                                  : moveColumnRight(header.column.getIndex()),
+                            },
+                            ...(header.column.getIsSorted() === 'asc'
```
john-groppe, line 489: There is no check if sorting is disabled for this column. So, sorting options always appear. Is that outside the scope of this initial drop?

design-toolkit/components/src/index.css
```diff
@@ -234,6 +234,14 @@
   --icon-color: --value(--color- *);
 }
 
+@custom-variant data-selected (&[data-selected~="true"]);
```
ArrayKnight, line 237: There's already a variant for selected. If there's an issue with the existing one then lets discuss what needs to be fixed

design-toolkit/components/src/index.css
```diff
@@ -234,6 +234,14 @@
...
+@custom-variant data-active (&[data-active~="true"]);
+@custom-variant data-pinned (&:is([data-pinned~="top"], [data-pinned~="bottom"]));
```
ArrayKnight, line 239: Let's remove the data-

We should make it clear whether this is a boolean and/or enum type of selector.

As an enum selector, you would have value specific selectors:
`@custom-variant pinned-top (&[data-pinned="top"])`

As a boolean selector, you would have:
`@custom-variant pinned (&[data-pinned])`

Also, thank you for teaching me the shorthand and not requiring the use of `@slot`. I can go back and simplify the other variant selectors

design-toolkit/components/src/index.css
```diff
@@ -234,6 +234,14 @@
   --icon-color: --value(--color- *);
 }
 
+@custom-variant data-selected (&[data-selected~="true"]);
+@custom-variant data-active (&[data-active~="true"]);
```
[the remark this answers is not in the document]
sotomaque: Screenshot 2025-07-21 at 9 31 36 PM


not quite the same thing as ":active" / pressed / selected; 

`data-active` is used when the kebab menu for a given row is open, setting the background of that row to the highlight color; happy to rename to `data-menu-open` or `data-actions-open` or whatever you like best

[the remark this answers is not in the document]
devindavies: This was a concept I went back and forth with design on, because they wanted this to be state representing that you're interacting with a single row (as opposed to multiple rows). So this doesn't necessarily have to be triggered. by the menu. They mentioned theoretical situations where clicking a row opens a modal.

shadyendless (approved):

design-toolkit/components/src/components/pagination/types.ts
```diff
@@ -0,0 +1,27 @@
...
+export type PaginationProps = RefAttributes<HTMLDivElement> &
+  HTMLAttributes<HTMLDivElement> &
+  VariantProps<typeof PaginationStyles> & {
+    onPreviousPage: () => void;
+    isPreviousPageDisabled: boolean;
+    onNextPage: () => void;
+    isNextPageDisabled: boolean;
+    pageCount: number;
+    pageIndex: number;
+    setPageIndex: (pageIndex: number) => void;
```
ArrayKnight, line 26: You could reduce all of these callbacks to a single callback with this signature and the name of onChange

design-toolkit/components/src/components/pagination/utils.test.ts
```diff
@@ -0,0 +1,33 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { pagination } from './utils';
+
+// Mock the constants module
+vi.mock('./constants', () => ({
```
ArrayKnight, line 17: Why are you mocking this to the same value? Shouldn't this be a prop anyways?

design-toolkit/components/src/components/table/static-table.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import type { Meta, StoryObj } from '@storybook/react';
+import { Table } from './index';
```
ArrayKnight, line 14: Please update any imports from `./index` to `./`

ArrayKnight (changes requested):
You'll need to reach out to Jacob (or get him to commit comments to this PR) about the feedback he's got from trying to implement this in NCCT

The addition of the "context menu" and pagination seems to not be in the scope of this PR

Also, be aware of what you're exporting. Notice that the indexer is picking everything up which may not be something that should actually be public or should be named more specifically to avoid confusing or conflict

design-toolkit/components/src/index.css
```diff
@@ -240,6 +240,13 @@
   --icon-color: --value(--color- *);
 }
 
+/* Table Related Custom Variants */
```
ArrayKnight, line 243: These should be mixed in with the other custom variants. The order in which they are in this file is important and dictates. Also, I had an open question about what `active` represents and if we already have a selector that aligns with its purpose

design-toolkit/components/src/components/pagination/utils.test.ts
sotomaque: didnt want the test to fail if the const ever changed

PR #331, opened 2025-07-03
Add drawer

jstdwt:
Closes [PATH-436](https://gohypergiant.atlassian.net/browse/PATH-436)

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

https://github.com/user-attachments/assets/402a2311-b25e-490f-9bec-f7d7943184cc

https://github.com/user-attachments/assets/11db05fa-b9a8-44d9-a9cb-608e3cc7a366

https://github.com/user-attachments/assets/cc6db34d-3c40-4f1e-8c81-2d9051ac848e

- [ ] Yes
- [ ] No

### DIFF
--- a/.changeset/chilled-houses-travel.md
+++ b/.changeset/chilled-houses-travel.md
@@ -1,0 +1,5 @@
+---
+"@accelint/design-toolkit": minor
+---
+
+Refactor SearchField to follow styling standards
--- a/design-toolkit/components/src/components/icon/styles.ts
+++ b/design-toolkit/components/src/components/icon/styles.ts
@@ -14,7 +14,7 @@ import { tv } from '@/lib/utils';
 
 export const IconStyles = tv({
   base: [
-    'block h-[var(--icon-size,var(--icon-size-m))] w-[var(--icon-size,var(--icon-size-m))] [color:var(--icon-color,currentColor)]',
+    'block h-[var(--icon-size,var(--icon-size-m))] w-[var(--icon-size,var(--icon-size-m))] flex-none [color:var(--icon-color,currentColor)]',
     'size-large:[--icon-size:var(--icon-size-l)]',
     'size-medium:[--icon-size:var(--icon-size-m)]',
     'size-small:[--icon-size:var(--icon-size-s)]',
--- a/design-toolkit/components/src/components/search-field/index.tsx
+++ b/design-toolkit/components/src/components/search-field/index.tsx
@@ -9,85 +9,128 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-
 'use client';
-import { cn } from '@/lib/utils';
-import {
-  CancelFill,
-  Loop as LoopIcon,
-  Search as SearchIcon,
-} from '@accelint/icons';
+
 import 'client-only';
-import { cva } from 'cva';
+import type { ProviderProps } from '@/lib/types';
+import { CancelFill, Loop, Search } from '@accelint/icons';
+import { createContext } from 'react';
 import {
   SearchField as AriaSearchField,
   Button,
+  type ContextValue,
   Input,
+  composeRenderProps,
+  useContextProps,
 } from 'react-aria-components';
 import { Icon } from '../icon';
+import { SearchFieldStyles, SearchFieldStylesDefaults } from './styles';
+import type { SearchFieldProps } from './types';
 
-export interface SearchFieldProps {
-  className?: string;
-  /** Disables the input field. */
-  isDisabled?: boolean;
-  /** Displays a loading spinner. */
-  isLoading?: boolean;
-  /** Displays placeholder text. */
-  placeholder?: string;
-  /** Whether the input has a filled background or outlined. */
-  variant?: 'filled' | 'outlined';
+const { field, input, search, loading, clear } = SearchFieldStyles();
+
+export const SearchFieldContext =
+  createContext<ContextValue<SearchFieldProps, HTMLDivElement>>(null);
+
+function SearchFieldProvider({
+  children,
+  ...props
+}: ProviderProps<SearchFieldProps>) {
+  return (
+    <SearchFieldContext.Provider value={props}>
+      {children}
+    </SearchFieldContext.Provider>
+  );
+}
+SearchFieldProvider.displayName = 'SearchField.Provider';
+
+/**
+ * SearchField - A customizable search input component built on React Aria Components
+ *
+ * Provides a search input with integrated search icon, loading state, and clear functionality.
+ * Supports two visual variants (filled/outlined), and granular styling control.
+ *
+ * @example
+ * // Basic search field
+ * <SearchField placeholder="Search..." />
+ *
+ * @example
+ * // Filled variant with custom styling
+ * <SearchField
+ *   variant="filled"
+ *   placeholder="Search products"
+ *   classNames={{
+ *     input: "bg-gray-100",
+ *     searchIcon: "text-blue-600"
+ *   }}
+ * />
+ *
+ * @example
+ * // With event handlers
+ * <SearchField
+ *   placeholder="Type to search"
+ *   onSubmit={(value) => console.log('Search:', value)}
+ *   onChange={(value) => setQuery(value)}
+ * />
+ *
+ * @example
+ * // Using context provider for default props
+ * <SearchField.Provider variant="filled">
+ *   <SearchField placeholder="Search 1" />
+ *   <SearchField placeholder="Search 2" />
+ * </SearchField.Provider>
+ */
+export function SearchField({ ref, ...props }: SearchFieldProps) {
+  [props, ref] = useContextProps(props, ref ?? null, SearchFieldContext);
+
+  const {
+    classNames,
+    inputProps,
+    isLoading = false,
+    variant = SearchFieldStylesDefaults.variant,
+    ...rest
+  } = props;
+
+  return (
+    <Icon.Provider size='small'>
+      <AriaSearchField
+        {...rest}
+        ref={ref}
+        className={composeRenderProps(classNames?.field, (className) =>
+          field({ className, variant }),
+        )}
+      >
+        <Icon className={search({ className: classNames?.search, variant })}>
+          <Search />
+        </Icon>
+        <Input
+          {...inputProps}
+          className={composeRenderProps(classNames?.input, (className) =>
+            input({ className, variant }),
+          )}
+          type='search'
+        />
+        {isLoading ? (
+          <Icon
+            className={loading({ className: classNames?.loading, variant })}
+          >
+            <Loop />
+          </Icon>
+        ) : (
+          <Button
+            className={composeRenderProps(classNames?.clear, (className) =>
+              clear({ className, variant }),
+            )}
+          >
+            <Icon>
+              <CancelFill />
+            </Icon>
+          </Button>
+        )}
+      </AriaSearchField>
+    </Icon.Provider>
+  );
 }
 
-const textFieldStyles = cva(
-  [
-    'hide-cancel block w-full rounded-round p-s pr-[30px] pl-[35px] font-display text-body-s outline',
-  ],
-  {
-    variants: {
-      isDisabled: {
-        true: 'text-disabled outline-interactive-disabled placeholder:text-disabled',
-        false:
-          'text-default-light placeholder:text-default-dark hover:outline-interactive-hover focus:outline-highlight',
-      },
-      variant: {
-        filled: 'bg-surface-raised outline-static-dark',
-        outlined: 'outline-interactive',
-      },
-    },
-  },
-);
-
-export const SearchField = ({
-  className,
-  placeholder = 'Search',
-  variant = 'outlined',
-  isLoading = false,
-  ...rest
-}: SearchFieldProps) => {
-  return (
-    <AriaSearchField className={cn('group relative', className)} {...rest}>
-      <Icon className='fg-interactive-hover absolute top-[6px] left-[7px]'>
-        <SearchIcon />
-      </Icon>
-      <Input
-        placeholder={placeholder}
-        className={({ isDisabled }) =>
-          cn(textFieldStyles({ isDisabled, variant }))
-        }
-      />
-      {isLoading ? (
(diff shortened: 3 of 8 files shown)


design-toolkit/components/src/components/drawer/styles.ts
```diff
@@ -0,0 +1,104 @@
...
+} as const;
+
+export const DrawerStyles = tv({
+  slots: {
+    overlay:
+      'absolute top-0 left-0 z-50 flex h-full w-full items-stretch justify-start duration-300 ease-in-out',
+    modal:
+      'relative flex h-full flex-col bg-surface-default text-body-m duration-300 ease-in-out',
+    dialog: 'flex flex-1 flex-col p-l text-body-m text-default-light',
+    attachedTrigger:
```
switzerb, line 29: this is massively nitpicky, but you can just name this `trigger` I think, without the `attached` description. I think it is already pretty clear what it refers to.

design-toolkit/components/src/components/drawer/styles.ts
```diff
@@ -0,0 +1,104 @@
...
+      'absolute z-60 transition-transform duration-300 ease-in-out',
+    header: 'mb-s flex flex-row items-center justify-between',
+    footer: 'mt-s flex flex-row items-center justify-end',
+    content: "flex-1"
+  },
+  variants: {
+    position: {
+      left: {
+        overlay: 'justify-start',
+        modal: 'w-[400px]',
```
switzerb, line 39: We are going to either want these to be configurable with a prop/css var or to respond to a min/max rule of some kind. The width may not be fixed.

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+  },
+};
+
+export const WithTabs: Story = {
+  render: () => {
+    return (
+      <div className='absolute inset-0 grid grid-cols-3 grid-rows-3 bg-default-light'>
+        <Tabs orientation='vertical'>
+          <Drawer position='left' isDismissable={false}>
+            <Drawer.Trigger attached>
```
switzerb, line 97: I'd like some docs here explaining in detail how to use the "attached" prop

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <Tabs.Panel id='b' className="h-full p-0">b content</Tabs.Panel>
+            </Drawer.Content>
+          </Drawer>
+        </Tabs>
+      </div>
+    );
+  },
+};
+
+export const WithNavigationStack: Story = {
```
switzerb, line 115: Couple of additional stories -- one with "shift content" behavior and the other with a custom button.

switzerb:
Will you please link to the Drawer ticket and add a short video capture and add it to the description above?

design-toolkit/components/src/components/drawer/index.tsx
```diff
@@ -0,0 +1,344 @@
+/*
```
switzerb, line 1: Since this is using context, you'll need a `'use client'` directive at the top and a `import 'client-only'`

design-toolkit/components/src/components/drawer/types.ts
```diff
@@ -0,0 +1,209 @@
...
+  className?: string;
+  children: ReactElement<DOMAttributes<FocusableElement>, string>
+  id?: MenuItemId;
+}
+
+export interface DrawerPanelProps extends ContainerProps {
+  id?: MenuItemId;
+}
+
+export interface UseDrawerToggleProps {
```
switzerb, line 106: This type appears to be unused

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+            {placeholderIcons}
+          </Drawer.Menu>
+
+          <PanelTitle>Top</PanelTitle>
+          {/* {LIPSUM} */}
+        </Drawer>
+
+        <Drawer.Main>
+          <div
```
switzerb, line 151: I appreciate the impulse to demonstrate the "overlay" behavior in the story, but the wallpaper background here is pretty aggressive. And I think the "Left Panel Content" method also feels a little confusing. 

CleanShot 2025-07-23 at 11 11 11@2x

It could be just personal preference, but the simple `lorem ipsum` like we have here in the previous iteration of the drawer feels more intuitive to get the point across.

CleanShot 2025-07-23 at 11 13 19@2x

design-toolkit/components/package.json
```diff
@@ -95,6 +98,7 @@
     "@accelint/biome-config": "workspace:*",
     "@accelint/typescript-config": "workspace:*",
     "@accelint/vitest-config": "workspace:*",
+    "@react-aria/interactions": "3.25.3",
```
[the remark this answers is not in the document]
ArrayKnight: That's actually not what we decided. Because of `pnpm`'s default behavior to add it to both and because it doesn't install peer deps by default, having it in both places is the way to go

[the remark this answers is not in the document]
switzerb: Out of curiosity, what is the advantage of not auto-installing peers?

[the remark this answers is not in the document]
ArrayKnight: It's just not the default behavior. So instead of implementing the workaround, we should embrace the way pnpm has decided that it's intended to be used

design-toolkit/components/src/components/drawer/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  return (
+    <DrawerLayoutContext.Provider value={drawerState}>
+      <div
+        className={root({ className: classNames?.layout })}
+        data-extend={extend}
+        {...dataAttributes}
+      >
+        {children}
+      </div>
+    </DrawerLayoutContext.Provider>
```
thebdewitt, line 74: @ArrayKnight I would like to use the provider pattern here with slots for header, title, footer, main, and content.  I noticed we head the generic Element in the design-system. Do you have a recommended approach?

ArrayKnight: No, I don't want to reintroduce that component. It was being abused / overused due to its generic nature. People were using it like a `div`

design-toolkit/components/src/components/drawer/types.ts
```diff
@@ -0,0 +1,186 @@
...
+  state: DrawerState;
+}
+
+/**
+ * Extended Drawer Layout Configurations
+ *
+ * The layout system supports four different drawer extension modes that determine
+ * how drawers are arranged and which drawers extend to the full container dimensions.
+ *
+ * extend: "tall"
```
switzerb, line 121: This is a little confusing since, according to the actual type below, I think this should say "left right"?

```suggestion
 * extend: "left right"
```

design-toolkit/components/src/components/drawer/types.ts
```diff
@@ -0,0 +1,186 @@
...
+ * ┌──────┬──────────┬───────┐
+ * │      │   top    │       │
+ * │      ├──────────┤       │
+ * │ left │   main   │ right │
+ * │      ├──────────┤       │
+ * │      │  bottom  │       │
+ * └──────┴──────────┴───────┘
+ *
+ * extend: "wide"
```
switzerb, line 130:
```suggestion
 * extend: "top bottom"
```

design-toolkit/components/src/components/navigation-stack/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  children: ReactNode;
+};
+
+export interface NavigationStackNavigateProps {
+  children: ReactElement<DOMAttributes<FocusableElement>, string>;
+  /**
+   * The string is to be used as a childId. When behavior is a childId
+   * navigate will push the childId onto the stack.
+   * */
+  behavior: LiteralUnion<'back' | 'clear', string>;
```
switzerb, line 39: nit: calling this `behavior` feels somewhat confusing to me. maybe something more directly informative like `to` or `action`?

design-toolkit/components/src/components/navigation-stack/navigation-stack.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const Default: Story = {
+  render: (args: NavigationStackProps) => (
+    <div className='h-[300px] w-[200px] text-default-light'>
+      <NavigationStack defaultViewId='parent' {...args}>
+        <NavigationStack.View id='parent'>
+          <div className='flex flex-col gap-m'>
+            <div className='text-header-l'>Parent View</div>
+            <NavigationStack.Navigate behavior='child'>
+              <span className='text-body-s'>Go To Child</span>
```
switzerb, line 24: nit: make this "Go To Child" a button so it is clear that it is a click behavior. In the story, it is a little hard to tell what is going on. Also, since this is a new component with a pretty abstract purpose, I'd like to see even a minimal one or two sentence description of what it is and where is it used.

design-toolkit/components/src/components/navigation-stack/types.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface NavigationStackContextValue {
+  currentViewId: string | null;
```
switzerb, line 43: I think these ids should be `Key` types instead of `strings` to be more flexible.

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,443 @@
...
+  component: Drawer,
+  parameters: {
+    layout: 'fullscreen',
+  },
+  args: {
+    id: 'left-drawer',
+    placement: 'left',
+    size: 'medium',
+    extend: 'left right',
+    push: 'left',
```
switzerb, line 42: Please provide a select control for size, push, placement, etc. Also, since the "extends" really only applies to the "multiple drawers" scenario, I would include the "extend" args only on that story since otherwise they look like they don't do anything and the name is not immediately intuitive. I LOVE the code comments in the types though. That is awesome.

CleanShot 2025-07-30 at 12 25 51@2x

design-toolkit/components/src/components/drawer/types.ts
thebdewitt: good call - I was experimenting with some ideas here

design-toolkit/components/src/components/navigation-stack/types.ts
thebdewitt: Both of those options resonate w/me.  @ArrayKnight suggested behavior.  Would `to` or `action` cause collisions elsewhere? We also introduced `behavior` on `Drawer.Trigger`.  We should land on something that works in both places.

design-toolkit/components/src/components/navigation-stack/navigation-stack.stories.tsx
thebdewitt: I wanted to make sure `NavigationStack.Navigate` works with a non-clickable element as well.

design-toolkit/components/src/components/navigation-stack/types.ts
switzerb: Oh yeah, `for` is more intuitive, too. It's a minor usability nit so it's not a blocker for me.

design-toolkit/components/src/components/drawer/drawer.stories.tsx
thebdewitt: I created a separate for `Drawer.Layout`.  That was driving me crazy to see the props and they don't do anything. I don't know of a better way to determine props per story and have a clean separation.  

I think push should be a text input, because it is a combination of values that is quite large.  

I can't take credit for the code comments, that is @kalisjoshua 's work :smile: I can add some more though!

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+            </Drawer>
+          </Drawer.Layout>
+        </Drawer.Provider>
+      </div>
+    );
+  },
+};
+
+export const WithLongContent: Story = {
+  render: () => {
```
switzerb, line 90: Need to pass args from the Storybook controls into the component.
```suggestion
  render: (args) => {
```

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+    return (
+      <div className='h-screen w-full'>
+        <Drawer.Provider>
+          <Drawer.Layout>
+            <Drawer.Main>
+              <div className='h-dvh bg-surface-raised' />
+            </Drawer.Main>
+            <Drawer id='settings' placement='left'>
```
switzerb, line 98:
```suggestion
            <Drawer id='settings' placement='left' {...args}>
```

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+              <Drawer.Menu>
+                <Drawer.Menu.Item>
+                  <Placeholder />
+                </Drawer.Menu.Item>
+              </Drawer.Menu>
+              <Drawer.Content>
+                <Drawer.Header>
+                  <Drawer.Title>Title</Drawer.Title>
+                  <Drawer.Trigger for='settings' behavior='close'>
+                    <Button size='small'>
```
switzerb, line 108: The inconsistency with the Button styles here for closing is a little confusing. I would stick to just the "flat" variant so that it matches the Core design system

design-toolkit/components/src/components/drawer/drawer.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+            </Drawer>
+          </Drawer.Layout>
+        </Drawer.Provider>
+      </div>
+    );
+  },
+};
+
+export const Uncontrolled: Story = {
+  render: () => {
```
switzerb, line 181: ...same args comment here.

switzerb (approved):
Nice work! Found a few small things that can be left to a fast-follow ticket so we can get this merged in.

PR #334, opened 2025-07-04
feat: menu component

switzerb:
Closes https://gohypergiant.atlassian.net/browse/PATH-320

https://github.com/user-attachments/assets/00ed902b-388c-4e0f-ade2-850f1ee2e7ab

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run the storybook instance with `pnpm --filter=@accelint/design-toolkit preview` to view the Menu stories

- [ ] Yes
- [x] No

### DIFF
--- a/design-toolkit/components/src/components/menu/menu.test.tsx
+++ b/design-toolkit/components/src/components/menu/menu.test.tsx
@@ -15,8 +15,8 @@ import Placeholder from '@accelint/icons/placeholder';
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { describe, expect, it } from 'vitest';
+import { Button } from '../button';
 import { Icon } from '../icon';
-import { IconButton } from '../icon-button';
 import { Menu } from './index';
 import type { MenuProps } from './types';
 
@@ -83,11 +83,11 @@ function setup({
 }: Partial<MenuProps<object>> = {}) {
   render(
     <Menu.Trigger>
-      <IconButton aria-label='Menu'>
+      <Button variant='icon' aria-label='Menu'>
         <Icon>
           <Kebab />
         </Icon>
-      </IconButton>
+      </Button>
       <Menu {...rest}>{children}</Menu>
     </Menu.Trigger>,
   );


design-toolkit/components/src/components/menu/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  MenuTextProps,
+  SeparatorProps,
+} from './types';
+
+const { item, label, description, menu, more, icon, separator } = MenuStyles();
+
+export const MenuContext =
+  createContext<ContextValue<MenuProps<unknown>, HTMLDivElement>>(null);
+
+export function Menu<T extends object>(props: MenuProps<T>) {
```
ArrayKnight, line 47: The menu should be using the menu context as well in case a parent is dictating props

design-toolkit/components/src/components/menu/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+export const MenuItem = (props: MenuItemProps) => {
+  const context = useContext(MenuContext);
+  const variant =
+    (isSlottedContextValue(context) ? undefined : context?.variant) ??
+    MenuStylesDefaults.variant;
+
+  const { children: childrenProp, className, ...rest } = props;
+
+  const children = useCallback(
```
ArrayKnight, line 84: If we're relying on the React compiler (as I've been told by Jacob), we shouldn't need the useCallback

design-toolkit/components/src/components/menu/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+      {children}
+    </AriaMenuItem>
+  );
+};
+
+export function MenuSection<T extends object>(props: MenuSectionProps<T>) {
+  const { header, children, items } = props;
+
+  return (
+    <AriaMenuSection id={header}>
```
ArrayKnight, line 121: Is an id required?

design-toolkit/components/src/components/menu/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  TextProps as AriaTextProps,
+} from 'react-aria-components';
+import type { IconProps } from '../icon/types';
+import type { MenuStyleVariants } from './styles';
+
+type BaseProps = {
+  className?: string;
+};
+
+export type MenuProps<T> = Omit<AriaMenuProps<T>, 'className'> &
```
ArrayKnight, line 28: I don't think that we should be removing access to the className render props

design-toolkit/components/src/components/menu/index.tsx
brandonjpierce: JERIC2O will be a little delayed in adopting the compiler if I had to guess so not a massive harm in including the `useCallback` here. Eventually once all projects are using the compiler we can come back through and remove. We would also need to test the compiler output against this component to see if it is correctly optimizing.

switzerb:
> The linked issue shows a screenshot of Figma, but the linked Figma seems to be slightly different. I bring this up because I wanted to check the intended spacing. What I'm seeing in the video you've attached is that the compact size doesn't affect the spacing between the label and the description which surprised me. It seems too large considering the other spacing reduction occuring

The figma file needs to be brought up to date to make it more clear, but the screenshot in the ticket is for the "Options" component which is also being used here for the menu: https://www.figma.com/design/yMWadmpVp2WC6VAOpFQwqx/CORE-Design-System?node-id=5402-8762&t=efsC73VrcBvMnAqL-0

And according to the Options, the gap stays the same and so does the label/description relationship (and the icon size). You think it is worth a conversation with design?

design-toolkit/components/package.json
```diff
@@ -8,9 +8,7 @@
     "url": "https://github.com/gohypergiant/standard-toolkit/design-toolkit"
   },
   "type": "module",
-  "files": [
-    "dist/**"
-  ],
+  "files": ["dist/**"],
```
switzerb, line 11: the flip-flopping on this formatting is making me crazy. we clearly have conflicting rules somewhere.

design-toolkit/components/src/components/menu/menu.stories.tsx
```diff
@@ -0,0 +1,230 @@
...
+        id: 2,
+        prefixIcon: <Placeholder />,
+        name: 'Blue jay',
+        description: 'Cyanocitta cristata',
+        color: 'serious',
+      },
+      {
+        id: 3,
+        prefixIcon: <Placeholder />,
+        name: 'Gray catbird',
```
gromkii, line 69: Solid bird choices.

gromkii (approved):
Awesome work. 🙌

orteth01:
i tested locally and it generally looks great!

i think a worthwhile story/test would be one to ensure that the menu behaves correctly when it hits a viewport boundary. this is important for context menus on the cop, where users might right click close to the viewport boundary

example from amorphous (plz ignore our custom styles)
Screenshot 2025-07-18 at 10 55 24 AM
Screenshot 2025-07-18 at 10 55 37 AM




in this video, you can see how the default react-aria-component menu works at viewport boundary vs how our stories currently handle it.

https://github.com/user-attachments/assets/3fa06b6f-ac9f-425d-855a-bec33ad93f50

PR #340, opened 2025-07-10
Refactor/chip

natalieyip:
Closes 

## ✅ Pull Request Checklist:

- [ ] Yes
- [ ] No

### DIFF
--- a/design-toolkit/components/src/components/chip/index.tsx
+++ b/design-toolkit/components/src/components/chip/index.tsx
@@ -12,7 +12,6 @@
 
 import { cn } from '@/lib/utils';
 import { CancelFill } from '@accelint/icons';
-import { type VariantProps, cva } from 'cva';
 import type React from 'react';
 import { type ReactNode, createContext, useContext } from 'react';
 import {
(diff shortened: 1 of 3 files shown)


design-toolkit/components/src/components/chip/styles.ts
```diff
@@ -0,0 +1,70 @@
...
+      small: 'px-s py-xs text-body-xs',
+    },
+  },
+  defaultVariants: {
+    size: 'medium',
+    variant: 'info',
+  },
+});
+
+export const SelectableChipStyles = tv({
```
ArrayKnight, line 36: Would it be possible to use the extend feature of TV to use the base styles and only add on the delta?

ArrayKnight (changes requested):
Good first pass

The other thing I'm going to look into is if we can get rid of the Coordinator

ArrayKnight:
To get rid of the Coordinator context, I believe that you can instead detect the presence of the `TagListContext`: https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/TagGroup.tsx#L111

design-toolkit/components/src/components/chip/index.tsx
```diff
@@ -163,43 +110,17 @@ export const SelectableChip = ({
...
 }: SelectableChipProps) => (
   <AriaTag
-    className={cn(
-      selectableChipStyles({
-        isDisabled,
-        size,
-        className,
-      }),
+    className={composeRenderProps(className, (className) =>
+      SelectableChipStyles({ size: size, isDisabled: isDisabled, className }),
```
sotomaque, line 114:
```suggestion
      SelectableChipStyles({ size, isDisabled, className }),
```

design-toolkit/components/src/components/chip/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  ChipStyles,
+  DeletableChipStyles,
+  SelectableChipStyles,
+} from './styles';
+
+export type ChipProps = VariantProps<typeof ChipStyles> &
+  Omit<HTMLProps<HTMLSpanElement>, 'children' | 'size'> & {
+    className?: string;
+    /** Used to add text to the badge, such as the number of unread notifications. */
+    children?: ReactNode;
```
ArrayKnight, line 30: No need to redefine children. You can use the children already on the above type

design-toolkit/components/src/components/chip/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+export type ChipListProps<T> = Omit<AriaTagGroupProps, 'children'> &
+  Pick<AriaTagListProps<T>, 'items' | 'children' | 'renderEmptyState'>;
+
+export type SelectableChipProps = VariantProps<typeof SelectableChipStyles> &
+  Omit<AriaTagProps, 'isDisabled'>;
```
ArrayKnight, line 37: Is there a reason for this omit?

ArrayKnight:
One thing I just noticed is that the Storybook stories seem to missing the concept of the disabled state

Also the SelectableChipStyles should have the isSelected variant instead of using the dtk-selected class

ArrayKnight:
@natalieyip FYI, my guidance on getting rid of the context was invalid. Due to complexities with RAC's collection renderer, the contexts that I said you could test against are not available in that rendering scope. I've updated the existing context implementation to have a more descriptive name to describe it's purpose.

design-toolkit/components/tsconfig.json
```diff
@@ -10,6 +10,7 @@
   ],
   "compilerOptions": {
     "jsx": "react-jsx",
+    "lib": ["ESNext", "DOM"],
```
ArrayKnight, line 13: This was necessary to be able to use Set methods, but should ultimately be handled in the tsconfig package

PR #389, opened 2025-07-29
Fix: esbuild output

brandonjpierce:
If a path alias was used in tsconfig `esbuild` would generate an incorrect path import in the build output. E.g. 

```import { } from '@/thing'``` 

would produce 

```import { } from './../../thing.js';``` 

instead of 

```import { } from './../../thing/index.js';```

This was due to one of our esbuild plugins eagerly attaching the file extension before the path alias resolution took place. It turns out one of our other plugins can actually do the job of appending the file extension (to make it ESM spec compliant) so it was easy as removing a plugin.

This also fixes an issue where we had multiple versions of `esbuild` in the repo.

## 📝 Test Instructions:

`pnpm run build`

Analyze the output in the `dist` folder to ensure things were appropriately produced.

- [ ] Yes
- [x] No

## 💬 Other information

Per the instructions in the plugin, `bundle` should be set to `false` in esbuild to ensure appropriate compilation.

### CODE REVIEW

**Review by devindavies — APPROVED** [review]

**Review by ArrayKnight — DISMISSED** [review]

File: design-toolkit/components/src/components/chip/types.ts [hunk 1]
```diff
@@ -33,7 +33,7 @@ export type ChipListProps<T> = Omit<TagGroupProps, 'children'> &
   BaseChipProps;
 
 export type ChipProps = VariantProps<typeof ChipStyles> &
-  Omit<ComponentPropsWithRef<'div'>, 'size'> &
+  Omit<ComponentPropsWithRef<'div'>, 'size' | 'onClick'> &
```
brandonjpierce, line 36: Sanity check this is intended

isaackehle: And, this is because we want the `OnClick` from the `BaseChipProps`, correct?

packages/design-system/src/components/menu/index.tsx
```diff
@@ -399,7 +399,7 @@ export const MenuItem = forwardRef(function MenuItem<T extends object>(
     <RACMenuItem<T>
       {...rest}
       id={id ?? textValue}
-      ref={ref as ForwardedRef<T>}
+      ref={ref}
```
isaackehle, line 402: 👍🏻

packages/design-system/src/components/tree/index.tsx
```diff
@@ -559,7 +559,7 @@ export function TreeItem<T>({
       ],
       // Need to expand the RAC ButtonContext to allow for additional slots
       [RACButtonContext, { slots: { [DEFAULT_SLOT]: {}, lines: {} } }],
-    ];
+    ] as TreeItemContexts;
```
isaackehle, line 562: nit: is there a way to do this without typecasting it? ie: building it above this with populating all required params?

design-toolkit/components/src/components/options/types.ts
```diff
@@ -27,10 +27,10 @@ export type OptionsDataItem = {
 };
 
 export type OptionsItemProps<T extends OptionsDataItem> = Omit<
-  ListBoxItemProps,
+  ListBoxItemProps<T>,
   'className'
 > &
-  RefAttributes<T> & {
+  RefAttributes<HTMLDivElement> & {
```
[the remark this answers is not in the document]
isaackehle: question: This seems to come out of left field. `RefAttributes` doesn't have a default type for `<T>` ?

isaackehle (commented):
a few questions for now. Otherwise, looks pretty good.

design-toolkit/components/src/components/chip/types.ts
brandonjpierce: I think this just came along when I merged in `main` to this branch so I don't know the full context of the change

design-toolkit/components/src/components/options/types.ts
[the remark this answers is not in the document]
brandonjpierce: Ditto above, this came from `main` so I am lacking the context of the change

packages/design-system/src/components/query-builder/value-selector.tsx
```diff
@@ -46,7 +46,9 @@ export function ValueSelector<T extends FullOption = FullOption>({
   });
 
   const handleSelectionChange = useCallback(
-    (selection: Key) => onChange(`${selection}`),
+    (selection: Key | null) => {
+      onChange(selection ? `${selection}` : []);
```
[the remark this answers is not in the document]
brandonjpierce: cc @ArrayKnight as I think he did the change

packages/design-system/src/components/tree/index.tsx
brandonjpierce: cc @ArrayKnight since he did this change

design-toolkit/components/src/components/chip/types.ts
ArrayKnight: This is ok. It came in with some the dep & ts updates. The props belong to a non-interactive element, and it's just a type deconfliction issue

design-toolkit/components/src/components/options/types.ts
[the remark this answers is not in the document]
ArrayKnight: This is okay too. RAC used to reference the underlying data `T` as the ref type, but it was in fact wrong. The ref actually always points at the element, in this case a `div`. With the update to RAC, this fix came through and we adjusted to match

ArrayKnight (dismissed):

ArrayKnight (approved):

Issue #267, opened 2025-04-30
React hook for @accelint/bus


brandonjpierce:
### Provide a general summary of the feature

Probably something along the lines of:

```
const emit = useEmit('test');

...

emit({ some: 'payload' });
```

```
// should auto cleanup under the hood
// need to make sure callback is stored appropriately
useListener('test', (e) => {
  // e.payload.some -> 'payload'
});

useListener('test', (e) => {
  // e.payload.some -> 'payload'
}, { once: true });
```

### 🤔 Expected Behavior?

-

### 📜 Context

-

PR #415, opened 2025-08-12
Feat: bus hooks

brandonjpierce:
ℹ️ Uses a ponyfill for `useEffectEvent`

Closes https://github.com/gohypergiant/standard-toolkit/issues/267

```
function MyComponent(props) {
  const { foo } = props;
  const [thing, setMyThing] = useState(false);

  const emit = useEmit('some-event');

  useOn('some-event', (payload) => {
    // this callback is stable and you can access props / state without
    // the values becoming stale. Event is automatically cleaned up.

    console.log(foo);
    console.log(thing);
    console.log(payload);
  });

  const onClick = () => {
    emit({ some: 'payload' })
  }

  return (
    <button onClick={onClick}>Fire Event</button>
  )
}
```

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test suite added, run `pnpm --filter=@accelint/bus test`

- [ ] Yes
- [x] No

### DIFF
--- a/.changeset/sweet-nails-boil.md
+++ b/.changeset/sweet-nails-boil.md
@@ -1,0 +1,5 @@
+---
+"@accelint/bus": minor
+---
+
+Create React hooks for events with type safety and render safety


packages/bus/package.json
```diff
@@ -40,14 +38,19 @@
...
     "esbuild-fix-imports-plugin": "1.0.21",
+    "react": "^19.0.0",
     "tsup": "8.5.0",
     "vitest": "2.1.3",
     "vitest-broadcast-channel-mock": "0.1.0"
   },
   "dependencies": {
     "typescript": "^5.8.3"
   },
+  "optionalDependencies": {
```
ArrayKnight, line 51: Why optional? Because most things in the package don't rely on it? And if you're using the React bits you probably already have React?

packages/bus/src/react/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import { useEffect } from 'react';
+import { Broadcast } from '../broadcast';
+import { useEffectEvent } from './ponyfill';
+
+const bus = Broadcast.getInstance();
+
+export function useEvent(event: any, callback: () => void) {
```
ArrayKnight, line 19: We'll need to improve these types, especially now the bus accepts generics

packages/bus/src/react/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  const onCallback = useEffectEvent(callback);
+
+  useEffect(() => {
+    const off = bus.on(event, onCallback);
+    return () => off();
+  });
+}
+
+// Alias
+export const useOn = useEvent;
```
ArrayKnight, line 29: Why do we need both?

packages/bus/src/react/index.ts
brandonjpierce: Yep I was waiting on the bus typescript branch to merge before I properly added any of the types here

packages/bus/src/react/index.ts
brandonjpierce: This was mostly an idea to match the actual method names from the `Broadcast` class itself e.g. `on` and `emit`, not 100% necessary

packages/bus/package.json
brandonjpierce: The react hook is opt in regarding usage, meaning you can install and use `@accelint/bus` without React. Putting `react` as an optional dependency reflects that.

PR #442, opened 2025-08-22
feat: implement light mode

orteth01:
Closes 

TODO:
- [ ] Test _everything_ in storybook. look for oddities
- [ ] Refine the ThemeProvider. It needs overrides still (and eventually extensions, but probably not in this PR)
- [ ] Update the next app to show example of implementing an override (accent color would be a good example)
- [ ] changeset (major bumpage)

## ✅ Pull Request Checklist

## 📝 Test Instructions

- run storybook: `pnpm --filter=@accelint/design-toolkit preview`
- click around and look for things that might be off

- [x] Yes
- [ ] No

- many utility classes that used to exist do not anymore. the utility classes for colors provided by the system are now constrained to match the tokens used in the design system
- 

## 💬 Other information

a script was created to help with this (design-toolkit/components/scripts/token-migration.mjs).
- it can be copied and used in other apps.
- the token swapping is a one off. it won't be super useful in the future.
- detecting color classes that aren't our system is potentially useful in the future - i hope instead the ecosystem improves and we get that for free from either the IDE extension or a linter. there's some stuff out there that worked with v3 of tailwind, but still doesn't work for v4.

### migration guide (what was done for this PR, but also what other apps can do)

- run script to replace utility classes
- the script will also highlight some classes it found that aren't in our theme system any more
	- these need to be changed manually
	- there may be some false positives. use good judgement.
- other flavors of classes that need to be manually looked at
	- text-*
		- previously we could use text-* classes for text color
		- should be changed to fg-*
	- border-*
		- previously we could muse border-* classes for border color
		- should be changed to outline-* for outlines
		- for things like dividers, probably change approach to use bg color??
	- potentially more....


design-toolkit/components/src/providers/theme-provider.tsx
```diff
@@ -0,0 +1,82 @@
...
+import type { PartialDeep } from 'type-fest';
+import { designTokens } from '../tokens/tokens';
+
+type Mode = 'dark' | 'light';
+
+type UseThemeContext = {
+  mode: Mode;
+  tokens: ColorTokens;
+  toggleMode: (mode: Mode) => void;
+} | null;
```
orteth01, line 35: i was thinking it might be nice to provide default values instead of null for the initial context. then the apps don't need to worry about null checks/optional chaining.

wdyt?

design-toolkit/components/src/index.css
```diff
@@ -24,6 +24,24 @@
...
+  --color-classification-ts-sci: var(--primitive-classification-ts-sci, #fce83a);
+  --color-roe-id-pending: var(--primitive-roe-id-pending, #404040);
+  --color-roe-id-friend: var(--primitive-roe-id-friend, #1484f4);
+  --color-roe-id-assumed-friend: var(--primitive-roe-id-assumed-friend, #0e8f37);
+  --color-roe-id-neutral: var(--primitive-roe-id-neutral, #5b137a);
+  --color-roe-id-unknown: var(--primitive-roe-id-unknown, #edda0a);
+  --color-roe-id-suspect: var(--primitive-roe-id-suspect, #ffad38);
+  --color-roe-id-hostile: var(--primitive-roe-id-hostile, #ff0033);
+  --color-roe-id-no-statement: var(--primitive-roe-id-pending, #404040);
+}
```
orteth01, line 43: i'm not sure why this was added here instead of adding semantic vars/utilities.

design-toolkit/components/scripts/generate-tokens.mjs
```diff
@@ -10,165 +10,42 @@
...
+  bg: ['background-color'],
+  fg: ['color', '--icon-color'],
+  outline: ['outline-color'],
+  shadow: ['box-shadow'],
+};
+const skipFallback = ['icon-size', 'shadow-elevation', 'font'];
 
-  // Return the matched variable name or null if not found
-  return varMatch ? varMatch[1] : null;
+//#region I/O utils
```
orteth01, line 32: i tried to organize things in this file a little bit
 
region comments make bigger files like this a little easier to navigate.
Screenshot 2025-08-22 at 3 31 40 PM

i think ideally, we'd move some of this stuff out to utils, and potentially even expose some things to apps (like the conversion utils).

also would be nice to write this in typescript eventually.

apps/next/src/app/kitchen-sink/page.tsx
```diff
@@ -0,0 +1,158 @@
...
+  Switch,
+  Tabs,
+  TextField,
+  useTheme,
+} from '@accelint/design-toolkit';
+import Placeholder from '@accelint/icons/placeholder';
+
+const divider = <div className='h-px w-full bg-accent-primary-muted' />;
+
+export default function KitchenSink() {
```
orteth01, line 31: just a nice little home for showing lots of UI things. i'm sure this will evolve in the future.

design-toolkit/components/src/index.css
switzerb: I left these here since they need to be available via the tailwind "magic" color compilation because, unlike the other semantic tokens, they can be used in background colors, foreground colors, etc. They also don't change with light/dark mode so it didn't make sense to include them as a part of that variant split.

design-toolkit/components/src/foundation/token-data.ts
```diff
@@ -0,0 +1,892 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// TODO: fine tooth comb
+// use this https://www.figma.com/design/yMWadmpVp2WC6VAOpFQwqx/CORE-Design-System?node-id=3-3040&m=dev&vars=1
+// it would be best if we could generate this somehow
```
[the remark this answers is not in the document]
switzerb: These drive the storybook stories. This was actually generated via the JSON/script but because I was in a big rush I didn't save the function that generates this data. It should really be generated as a part of that script (or if there is a better way to use the existing tokens to drive stories, that would work too. They are intended only for use internal to the stories which is why they have a `__private_exports` at the top there. Just for convenience of driving the visuals and printing the information for Storybook.

design-toolkit/components/src/components/drawer/styles.ts
```diff
@@ -135,14 +135,14 @@ export const DrawerMenuStyles = tv({
     ],
     item: [
       'flex flex-col items-center justify-center',
-      'fg-default-dark !p-xs cursor-pointer outline-none',
+      'sudo:fg-primary-muted !p-xs cursor-pointer outline-none',
```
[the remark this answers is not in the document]
orteth01: i agree. i opened a ticket for the !important usage last week - we probably should avoid using !important. but i wanted to limit scope to renaming classes that need to be renamed (and fixing any immediate oddities). 

i'm all for removing the important right now though 😅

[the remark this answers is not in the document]
orteth01: button has some fairly specific styles that need to be sudo'd often, i've found. w/out sudo this color did not get applied as expected.

design-toolkit/components/src/tokens/types.ts
```diff
@@ -0,0 +1,307 @@
...
+    'assumed-friend': RGBA;
+    neutral: RGBA;
+    unknown: RGBA;
+    suspect: RGBA;
+    hostile: RGBA;
+    'no-statement': RGBA;
+  };
+};
+
+export type SemanticColorTokens = {
```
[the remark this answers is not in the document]
orteth01: manually

design-toolkit/components/src/tokens/types.ts
```diff
@@ -0,0 +1,307 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// biome-ignore lint/style/useNamingConvention: acronym
+type RGBA = [number, number, number, number];
+
+export type StaticColorTokens = {
```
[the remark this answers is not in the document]
orteth01: it could! this is a very fresh thing added as an alternative to using the tw color magic for classification/roe colors.

the initial implementation of the theme provider seemed to be focused on providing colors, and that has been maintained. static colors were kept separate from other static properties as a side effect of that, i guess.

brenna and i did discuss if the theme provider should provide all tokens a little bit yesterday - do you have a preference on that?

.changeset/polite-pears-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/design-toolkit": major
+---
+
+Token overhaul. Add light mode.
```
switzerb, line 5: Since this is a breaking change and presumably devs and design are actually reading the changelog (hopefully?), I'd like to make this one a little more verbose:

```suggestion
Major theme updates --
* Adding theme support for light mode
* Renaming selected design tokens for light mode compatibility with
* Script for token naming migrations
* Default accent color update from seafoam green to accelint skyfoam blue
* Shift to using tailwind utility classes for color to better align with token names
```

orteth01: totally. makes sense

design-toolkit/components/scripts/token-migration.mjs
```diff
@@ -0,0 +1,197 @@
...
+  ['border-info', 'outline-info-bold'],
+  ['border-advisory', 'outline-advisory-bold'],
+  ['border-normal', 'outline-normal-bold'],
+  ['border-serious', 'outline-serious-bold'],
+  ['border-critical', 'outline-critical-bold'],
+]);
+
+const leGlob = 'src/**/*.{ts,tsx}';
+console.log(`🔍 Finding and replacing in ${leGlob}`);
+console.log();
```
colehart, line 151: lots of console logs in here that seem helpful for a script output except for this empty lil guy

design-toolkit/components/src/components/drawer/styles.ts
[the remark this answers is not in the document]
thebdewitt: I experienced padding not being applied for the icon variant on button, but after pulling in #446 it was good

PR #467, opened 2025-09-11
Path 366 storybook enhancements

kalisjoshua:
Closes PATH-266

1. Standardized (made consistent) the initialization of the story "meta": args, argTypes (controls), and parameters specifically
2. Create an audit script to check that standards are being followed in stories
    - This was largely created by copilot so either don't review that code until we decide to keep it
    - If there is better tooling out there I'm happy to evaluate but I didn't want to bring in ESLint just for this
3. Added the audit to the CI pipeline and lefthook pre-commit checks

I still need to work on:

- Component "actions"
- Test integration

## ✅ Pull Request Checklist

- [ ] Yes
- [x] No


devindavies:
Rather than an audit script, could we achieve something similar with [Biome plugins](https://biomejs.dev/linter/plugins/)?

It would be nice not have to parse the AST ourselves, maintain these audit scripts, and add a step to our CI pipeline

kalisjoshua:
> So, I can appreciate what you are trying to get done here -- there is a lot of repetition in our stories and certainly some inconsistency, both of which are problems worth solving in the engineering space.
> 
> However, from my perspective, the point of the Storybook stories we write are to communicate with our "customers" -- the other devs that we are supporting -- and to drive quality control (i.e. reduce regressions). These changes feel over-engineered and like they don't serve either of those two goals. More, it introduces a layer of indirection that makes the source a little harder to understand (at least for me) as well as adding a brittleness and inflexibility with the audit concept. To me, this feels like moving in the wrong direction.
> 
> However! -- I am open to learning that I am wrong. I think this would be a good topic for the Core Internal Sync so we can talk as a group. You can explain the thinking behind the engineering choices here and it would be good to get additional perspectives, I think, rather than just mine.

I have truncated the changeset drastically.

Most of the utilities are "opt-in", so stories can be written without them. The only one that is, currently, required is `createParameters()` to ensure a layout is chosen and some default values for `parameters.controls.exclude`; to reduce repetition.

Questions:

1. Do you not want an audit to ensure consistency of stories?
2. Do you not want standardized controls (`argTypes`)?

@switzerb

PR #477, opened 2025-09-16
chore: refresh syncpack update config file and versions

isaackehle:
Closes one link in a chain... 

There's not currently a Jira OR Github issue to link to. @switzerb if you have one, lmk! 

The goal of this is to freshen up the package versions across all of the output packages where determination safely can for now.  Afterwards, specific outlier packages can be upgraded. 

This is a prequel to adding in a basic scaffolding script that will use `syncpack` to ensure the packages are aligned properly with the rest of the monorepo.

I chose to relax the restriction on pinning versions of most packages, as in general when this happens, it becomes much harder to upgrade down the road, and there should be good reasons why pinning a version is happening. (ie what problem is being solved by that action?). This is up for debate of course, and hopefully we will have that discussion.

## ✅ Pull Request Checklist

- [ ] ~~Included link to corresponding [GitHub Issue](https://github.com/gohypergiant/standard-toolkit/issues).~~
- [ ] ~~Added/updated unit tests and storybook for this change (for bug fixes / features).~~
- [ ] ~~Added/updated documentation (for bug fixes / features)~~
- [ ] ~~Filled out test instructions.~~

- [ ] Yes
- [x] No

### DIFF
--- a/.changeset/neat-penguins-judge.md
+++ b/.changeset/neat-penguins-judge.md
@@ -1,0 +1,29 @@
+---
+"@accelint/constellation-tracker": patch
+"@accelint/design-toolkit": patch
+"@accelint/typescript-config": patch
+"@accelint/prettier-config": patch
+"@accelint/eslint-config": patch
+"@accelint/vitest-config": patch
+"@accelint/icons": patch
+"@accelint/map-toolkit": patch
+"@accelint/biome-config": patch
+"@accelint/converters": patch
+"@accelint/formatters": patch
+"@accelint/predicates": patch
+"@accelint/worker": patch
+"@accelint/constants": patch
+"@accelint/websocket": patch
+"@accelint/temporal": patch
+"@accelint/logger": patch
+"@accelint/smeegl": patch
+"@accelint/core": patch
+"@accelint/math": patch
+"@accelint/ntds": patch
+"@accelint/bus": patch
+"@accelint/geo": patch
+"@accelint/docs": patch
+"@apps/next": patch
+---
+
+Updated syncpack and realigned all packages for dependency versions


shadyendless (dismissed):

switzerb (dismissed):
LGTM

switzerb (dismissed):

brandonjpierce:
> Also, for all packages, we want to make sure we don't have any dependencies, only peer dependencies and dev dependencies.

I don't know if I necessarily agree with this as it puts a burden on the consumer if we blanket apply this rule to all of our packages. I think we should be selective in our `peerDependency` usage so long as we understand the downstream side effects for the consumer as well as the module graph in various situations (e.g. monorepo vs. not, etc).

isaackehle:
@switzerb @ArrayKnight  Pushed updates, and added changeset update. Please feel free to merge over the next couple of days if you're good with it.

design-toolkit/components/src/tokens/tokens.ts
```diff
@@ -251,7 +251,7 @@ export const designTokens = {
       accent: {
         primary: {
           bold: [57, 183, 250, 1],
-          hover: [110, 209, 255, 1],
+          hover: [0, 79, 126, 1],
```
colehart: were these two changes intentional?

design-toolkit/components/src/tokens/themes.css
```diff
@@ -185,7 +185,7 @@
     --fg-inverse-muted: var(--primitive-neutral-400, #888a8f);
     --fg-disabled: var(--primitive-neutral-300, #c3c5c7);
     --fg-accent-primary-bold: var(--primitive-accent-primary-300, #39b7fa);
-    --fg-accent-primary-hover: var(--primitive-accent-primary-200, #6ed1ff);
+    --fg-accent-primary-hover: var(--primitive-accent-primary-700, #004f7e);
```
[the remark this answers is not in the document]
isaackehle: See below

colehart (approved):

package.json
```diff
@@ -16,15 +16,16 @@
...
     "clean:cov": "pnpm -r exec rimraf coverage && exec pnpm rimraf coverage",
     "clean:deps": "pnpm -r exec rimraf node_modules && pnpm exec rimraf node_modules",
     "clean:dist": "pnpm -r exec rimraf dist",
-    "clean:buildinfo": "pnpm exec find . -name 'tsconfig.tsbuildinfo' -delete",
     "clean:pnpm": "pnpm rimraf pnpm-lock.yaml",
     "clean:turbo": "pnpm -r exec  rimraf .turbo && exec pnpm rimraf .turbo",
     "constellation-tracker": "turbo constellation-tracker",
     "deps:format-packages": "pnpm syncpack format",
     "deps:mismatch-fix": "pnpm syncpack fix-mismatches",
+    "deps:postinstall": "pnpm run deps:mismatch-fix && pnpm run deps:format-packages",
```
isaackehle: Note: in practice, this is **NOT** currently working. We will have to circle back to find out why this is the case. The idea was that once someone adds or updates a package in one of the projects, the `deps:postinstall` script is run to fix any mismatches, and then to format the `package.json` file. 

I think this deserves to be in the MR pipeline, and if anything changes, cause the pipeline to fail. That will cause the developer to look into the root cause and ask the questions.

design-toolkit/components/package.json
```diff
@@ -61,58 +60,28 @@
   "scripts": {
     "bench": "pnpm vitest bench --run --dir src",
     "build": "pnpm generate && pnpm license && pnpm format && tsup && pnpm build:css",
+    "build-storybook": "pnpm storybook build",
     "build:css": "pnpm tailwindcss -i ./src/index.css -o ./dist/styles.css",
     "constellation-tracker": "constellation-tracker",
     "dev": "pnpm tsc --watch",
+    "format": "pnpm biome format . --write",
```
switzerb:
```suggestion
    "format": "pnpm biome check . --write",
```

Based on earlier conversations...

PR #488, opened 2025-09-19
fix: update imports

brandonjpierce:
Audited the design toolkit code to ensure that `use client` and `import 'client-only';` were being appropriately added when necessary. We were missing the import for the classification banner/badge components.

Also ensured that our tailwind variant usage was standardized.


design-toolkit/components/src/components/action-bar/styles.ts
```diff
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { tv } from 'tailwind-variants';
+import { tv } from '@/lib/utils';
```
brandonjpierce, line 13: cc @switzerb @ArrayKnight wanted to make sure this change is okay. I noticed quite a few components were using `tailwind-variants` directly instead of our `@/lib/utils` `tv` wrapper.

brandonjpierce: If this is intentional I can revert!

ArrayKnight: This is a good catch! We definitely want to be using our internal tv function

ArrayKnight (approved):

PR #492, opened 2025-09-23
feat: add notice component

jstdwt:
Closes https://gohypergiant.atlassian.net/browse/PATH-309

## ✅ Pull Request Checklist

## 📝 Test Instructions
Open storybook and test the Notice and Notice.List stories

- [ ] Yes
- [x] No


design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,286 @@
...
+  critical: 'critical',
+};
+
+export function Notice({
+  id,
+  color = 'info',
+  message,
+  primary,
+  secondary,
+  showClose,
```
switzerb, line 86: The ticket notes say that the icon is optional, so you need another boolean prop here to "showIcon" and make that a conditional.

.changeset/afraid-pears-repair.md
```diff
@@ -0,0 +1,15 @@
...
+
+The Notice.List component provides queue management functionality that allows developers to programmatically add and remove notices from the display. Notices can be automatically removed after a specified timeout period, or they can be manually dismissed by users. The system supports flexible dequeuing options, allowing notices to be removed based on their unique identifiers, target list, color, or metadata.
+
+For enhanced user experience, the component includes a "shouldCloseOnAction" feature that automatically dismisses notices when users interact with action buttons, and a "hideClearAll" option that allows developers to hide the clear all button when managing notice visibility.
+
+The components include styling flexibility through classNames props, portal-based rendering for proper positioning, and integration with the existing design system's button components for consistent visual appearance across the application.
```
switzerb, line 15: I still trying to wrap my head around the guidance for changeset notes -- so I might be off-base here -- but I'm thinking one paragraph is more than sufficient to explain the component we are adding? You might be able to drop in a relative link to the story docs, too, for more info something like `[Notice](?path=/docs/components-notice--playground)`

ArrayKnight: Yea, I think I need to revise the guidance a bit. I didn't intend to introduce book writing to everyone's job description

.changeset/afraid-pears-repair.md
```diff
@@ -0,0 +1,6 @@
+---
+"@accelint/design-toolkit": minor
+---
+
+This release introduces a complete Notice component that provides developers with a comprehensive solution for displaying notifications, alerts, and messages. The implementation includes an individual Notice component for single notifications and Notice.List for queue-based notification management. The system supports five color variants (info, advisory, normal, critical, and serious) with customizable action buttons, automatic timeout management, flexible dequeuing options, and portal-based rendering. View the [Notice component in Storybook](?path=/docs/components-notice--playground) for interactive examples and documentation.
```
ArrayKnight, line 5: You'll need to make this URL absolute

switzerb: ah -- right -- it's a changeset and not a story. it won't be live until the work is merged -- but it is probably okay to guess? uncertain if that is clever or not.

switzerb:
There isn't a specific design for the "small" variant that I can find in Figma so this may not be problematic -- but the close button is positioned a little oddly here:

CleanShot 2025-09-26 at 09 23 04@2x

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+          {({ toast }: { toast: QueuedToast<NoticeContent> }) => (
+            <Notice
+              {...toast.content}
+              classNames={classNames?.notice}
+              size={size}
+              onAction={() => emitAction({ id: toast.key })}
+              onPrimaryAction={() => emitActionPrimary({ id: toast.key })}
+              onSecondaryAction={() => emitActionSecondary({ id: toast.key })}
+              onClose={() => {
+                toast.onClose?.();
```
ArrayKnight, line 286: Let's validate that the call to `queue.close` is necessary, it might already be taken care of by the `toast.close`

design-toolkit/components/src/components/notice/types.ts
```diff
@@ -0,0 +1,121 @@
...
+    id?: UniqueId;
+  }
+>;
+
+export type DequeueId = { id: UniqueId };
+export type DequeueList = { target: UniqueId };
+export type DequeueColor = { color: NoticeColor };
+export type DequeueMetadata = { metadata: Record<string, unknown> };
+
+type DequeuePayload = DequeueId | DequeueList | DequeueColor | DequeueMetadata;
```
ArrayKnight, line 108: We're going to want to modify this type to allow for combinations of these properties. For example, if I want to dequeue all of the notices of a color in a specific list with some specific metadata

design-toolkit/components/src/components/notice/types.ts
```diff
@@ -0,0 +1,121 @@
...
+  color?: NoticeColor;
+  primary?: ActionButtonProps;
+  secondary?: ActionButtonProps;
+  timeout?: number;
+  target?: UniqueId;
+  metadata?: Record<string, unknown>;
+};
+
+export type NoticeIconProps = {
+  variant?: NoticeColor;
```
ArrayKnight, line 43: Let's continue to use the term color instead of variant

design-toolkit/components/src/providers/portal.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { UNSAFE_PortalProvider } from '@react-aria/overlays';
+import { useIsSSR } from '@react-aria/ssr';
+import {
+  type PropsWithChildren,
+  type RefObject,
+  useEffect,
+  useState,
+} from 'react';
+
+export function PortalProvider({
```
ArrayKnight, line 22: Looks like we have similar versions of this in Dialog and Tooltip, with minor differences. If we're going to add this as an abstraction we need to make sure that it's capable of fulfilling both of those needs

design-toolkit/components/src/components/notice/styles.ts
```diff
@@ -0,0 +1,41 @@
...
+      'color-advisory:bg-advisory-muted color-advisory:outline-accent-primary-bold',
+      'color-normal:bg-normal-muted color-normal:outline-normal-bold',
+      'color-serious:bg-serious-muted color-serious:outline-serious-bold',
+      'color-critical:bg-critical-muted color-critical:outline-critical-bold',
+    ],
+    content: [
+      'flex flex-1 items-center group-size-medium/notice:gap-l group-size-small/notice:gap-m',
+    ],
+    region: [
+      'absolute flex flex-col gap-y-s',
```
ArrayKnight, line 31: The region should only be absolute if placement has been set. By default there shouldn't be a placement and the list should render in place

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+  });
+
+  useEffect(() => {
+    queue.subscribe(() => {
+      setHasNotices(queue.visibleToasts.length > 0);
+    });
+  });
+
+  return (
+    <PortalProvider parentRef={parentRef}>
```
ArrayKnight, line 252: The Notice.List should not default to being portaled to the document root. By default it should render in place without absolute positioning. If the placement prop has been set, then it should portal like how you have it (default of document root, with parentRef override)

design-toolkit/components/src/components/notice/index.tsx
jstdwt: I double checked and it looks like it is necessary

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+  useOn(NoticeEventTypes.queue, (data) => {
+    if ((id && data.payload.target === id) || !id) {
+      queue.add(
+        {
+          ...data.payload,
+          id: data.payload.id || uuid(),
+          color: defaultColor || data.payload.color,
+        },
+        {
+          timeout: defaultTimeout || data.payload.timeout,
```
ArrayKnight, line 219: You should use `??` instead of `||` here in case the `defaultTimeout` is `0`

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+    }
+  });
+
+  useOn(NoticeEventTypes.dequeue, (data) => {
+    // @ts-expect-error : queue.queue exists, but is not currently documented
+    const dequeue = queue.queue.filter((toast: QueuedToast<NoticeContent>) => {
+      return matchesMetadata(data.payload, toast.content);
+    });
+
+    // @ts-expect-error : queue.queue exists, but is not currently documented
```
ArrayKnight, line 231: Can you consolidate the accessor to a single instance to minimize the number of `@ts-expect-error`s please?

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+  ...rest
+}: NoticeListProps) {
+  const queue = useMemo(
+    () => new ToastQueue<NoticeContent>({ maxVisibleToasts: limit }),
+    [limit],
+  );
+  const state = useToastQueue(queue);
+
+  const [hasNotices, setHasNotices] = useState(false);
+  const ref = createRef<HTMLDivElement>();
```
ArrayKnight, line 200: Can you update this to `useRef` please?

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+          timeout: defaultTimeout || data.payload.timeout,
+        },
+      );
+    }
+  });
+
+  useOn(NoticeEventTypes.dequeue, (data) => {
+    // @ts-expect-error : queue.queue exists, but is not currently documented
+    const dequeue = queue.queue.filter((toast: QueuedToast<NoticeContent>) => {
+      return matchesMetadata(data.payload, toast.content);
```
ArrayKnight, line 228: NIT: prefer inline return when no other logic is happening

design-toolkit/components/src/components/notice/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+      'color-critical:bg-critical-muted color-critical:outline-critical-bold',
+    ],
+    content: [
+      'flex flex-1 items-center group-size-small/notice:flex-wrap group-size-medium/notice:gap-l group-size-small/notice:gap-m',
+    ],
+    message: 'flex-1 group-size-small/notice:basis-10/12',
+    region: [
+      'flex flex-col gap-y-s',
+      'data-placement:absolute',
+      'data-[placement=top]:-translate-x-1/2 placement-top:top-0 data-[placement=top]:left-1/2',
```
ArrayKnight, line 34: Please use the `:placement-*` variant selectors

design-toolkit/components/src/components/notice/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+      'data-[placement=bottom]:-translate-x-1/2 placement-bottom:bottom-0 data-[placement=bottom]:left-1/2',
+      'placement-right:right-0 data-[placement=right]:top-1/2',
+      'placement-left:left-0 data-[placement=left]:top-1/2',
+    ],
+    list: ['flex flex-col gap-y-s'],
+    actions:
+      'flex flex-row-reverse items-center gap-s group-size-small/notice:grow-1 group-size-small/notice:basis-1 group-size-small/notice:justify-start',
+    closeContainer:
```
ArrayKnight, line 42: You should be able to remove the `order` since the hierarchy matches already

I'd like to see if we can potentially simplify and get rid of the container and just have the close button. Is the container necessary?

design-toolkit/components/src/components/notice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+    | 'bottom'
+    | 'bottom left'
+    | 'left';
+  limit?: number;
+  defaultColor?: NoticeColor;
+  defaultTimeout?: number;
+  hideClearAll?: boolean;
+  size?: 'small' | 'medium';
+  global?: boolean;
+  classNames?: {
```
ArrayKnight, line 63: Should any of these classNames be render props compatible?

design-toolkit/components/src/components/notice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+};
+
+export type NoticeQueueEvent = Payload<
+  typeof NoticeEventTypes.queue,
+  Omit<NoticeContent, 'id'> & {
+    id?: UniqueId;
+  }
+>;
+
+//export type DequeueId = { id: UniqueId };
```
ArrayKnight, line 103: Make sure to clean up when you're done

design-toolkit/components/src/components/notice/index.tsx
```diff
@@ -0,0 +1,316 @@
...
+      )}
+      data-placement={placement}
+      queue={queue}
+    >
+      {children}
+    </ToastRegion>
+  ) : (
+    <div
+      {...regionProps}
+      className={region({ className: classNames?.region as string })}
```
ArrayKnight, line 305: You're going to need to either limit this className to a string (can' be passing a functional call to this prop) or creating a discrimination union type that states that when global is true, that className is able to be a function

ArrayKnight (approved):

vertexguy:
This has some bugs yet.  (test message)

ArrayKnight:
> Also note that none of the Serious or Critical versions appear to be using the correct button sets for the primary, secondary, or X: image

Figma:
image

Storybook:
image

Also notice that the designs directly contradict your statement:

> the rule is that you either a button(s), or you have the X, never both per the design

vertexguy:
> Also notice that the designs directly contradict your statement:
> 
> > the rule is that you either a button(s), or you have the X, never both per the design

Yeah, that's a fair point.  This component appears to have relied on a few text snippets in the anatomy section to describe the rules but didn't have good examples provided in the guidance section.  I can see how the state layout would make this a bit confusing too.  We're showing all things at once to check the states but that can also cause confusion.  My apologies, I will try adjusting that layout to make it more clear.

vertexguy:
OK I've updated the figma to describe all the configurations with visual examples.  I think this is good to go when:

1. The buttons used are Core, follow the min width rule, and show a pressed state before triggering the event.
2. Serious and Critical button styles are used on all 3 buttons with the matching serious / critical notice types
3. The 3 button types (primary, secondary, x) can be configured to be visible or not (as already suggested by SB)

PR #501, opened 2025-09-29
Add Clock Component

Peechiz:
Closes https://gohypergiant.atlassian.net/browse/PATH-691

## ✅ Pull Request Checklist

- [ ] Yes
- [x] IDK 
- [ ] No

### DIFF
--- a/design-toolkit/components/src/components/clock/clock.stories.tsx
+++ b/design-toolkit/components/src/components/clock/clock.stories.tsx
@@ -52,6 +52,6 @@ const customFormatter = new Intl.DateTimeFormat('en-US', {
   hour12: false,
 });
 
-export const CustomFormat: StoryObj<typeof Clock> = {
+export const CustomFormat: Story = {
   render: ({ ...rest }) => <Clock {...rest} formatter={customFormatter} />,
 };


design-toolkit/components/package.json
```diff
@@ -107,6 +107,7 @@
     "@accelint/bus": "workspace:*",
     "@accelint/core": "workspace:*",
     "@accelint/icons": "workspace:*",
+    "@accelint/temporal": "workspace:*",
```
Peechiz, line 110: is it correct to list this as a dev dependency? that feels weird since I'm directly using it for this component...

design-toolkit/components/src/components/clock/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * <Clock options={{ dateStyle: "short" }} /> // <time>9/30/25, 15:54:14 UTC</time>
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#locale_options| DateTimeFormatOptions MDN}
+ */
+export function Clock({ ref, options, ...rest }: ClockProps) {
+  const formatter = useMemo(() => {
+    return !options
+      ? DEFAULT_FORMATTER
+      : new Intl.DateTimeFormat('en-US', { ...DEFAULT_OPTIONS, ...options });
+  }, [options]);
```
Peechiz, line 54: memoizing this because the actual creation of the formatter is a little bit expensive (there are a lot of languages in there) and we want to avoid recreating it every tick.

Opted for a partial spread overwrite with the options so you don't have to specify _all_ the options if all you want to do is add a `dateStyle`, but still obviously want to use UTC.

Peechiz: Gippity said I could do something like `const stableOptions = useMemo(() => options, [JSON.stringify(options)]);` and then use `stableOptions` in the formatter memo instead.

design-toolkit/components/src/components/clock/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * <Clock className="fg-accent-primary-bold" />
+ *
+ * @example
+ * // Custom Format
+ * <Clock options={{ dateStyle: "short" }} /> // <time>9/30/25, 15:54:14 UTC</time>
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#locale_options| DateTimeFormatOptions MDN}
+ */
+export function Clock({ ref, options, ...rest }: ClockProps) {
+  const stableOptions = useMemo(() => options, [JSON.stringify(options)]);
```
ArrayKnight, line 50: What is happening here?

Peechiz: see above ⬆️

ArrayKnight: Sorry for missing that. I think I skipped right over it because it was labeled out of date

design-toolkit/components/src/components/clock/index.tsx
Peechiz: the other out-of-the-box approach here would be to turn all the `DateTimeFormatOptions` keys into _props_, and then populating the default values of the ones we're using instead, but that creates a :new: and different sort of mess.

ArrayKnight: The formatter wouldn't be recreated on every tick unless the prop / dependency array is provided a new instance of the options object

JSON.stringify is also expensive, and is being done for every tick / render

We could also potentially have it so we have them pass the entire, configured formatter instead of the options, since you're also hardcoding the locale

Peechiz: THE WHOLE FORMATTER! That's so much better.

ArrayKnight:
Looks good overall. You'll want to create the `clock.docs.mdx` file for consistent documentation within Storybook

design-toolkit/components/src/components/clock/clock.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+  },
+} satisfies Meta<typeof Clock>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  render: Clock,
+};
```
switzerb, line 40: optional: it might be confusing for devs when first loading this story because it is black, unstyled text on a default dark background. you can fix this by overriding the theme decorator, something like this:

```suggestion
export const Default: Story = {
  render: Clock,
  decorators: [
    withThemeByClassName<ReactRenderer>({
      themes: {
        light: 'light !bg-surface-default',
        dark: 'dark !bg-surface-default',
      },
      defaultTheme: 'light',
    }),
  ],
};
```

then again, we specifically say it is *unstyled* text so I can be convinced it is unnecessary.

switzerb (approved):
LGTM. One optional nit - and the docs may end up being too light but I like the simplicity.

design-toolkit/components/src/components/clock/clock.stories.tsx
Peechiz: this change caused some weird behaviors with light/dark mode behavior, so we're skipping it.

PR #504, opened 2025-09-30
refactor: updates deletable chip state styles

seeddisperser:
Updates the state styles of the delectable chip and modifies the hover, focus-visible and pressed behaviors to match the interactions from the design.


design-toolkit/components/src/components/chip/styles.ts
```diff
@@ -64,11 +64,20 @@ export const DeletableChipStyles = tv({
...
-      'gap-xs outline-interactive',
-      'enabled:focus-visible:outline-interactive-hover',
-      'enabled:hover:outline-interactive-hover',
-      'disabled:fg-disabled disabled:cursor-not-allowed disabled:outline-interactive-disabled',
+      'gap-xs outline-interactive cursor-default',
+      'enabled:fg-primary-muted',
+      'disabled:fg-disabled disabled:cursor-not-allowed disabled:outline-interactive-disabled disabled:bg-interactive-disabled',
+      'has-[.remove:focus-visible]:fg-hover',
+      'has-[.remove:hover]:fg-hover',
+      'has-[.remove[data-pressed]]:fg-pressed',
```
ArrayKnight, line 72: You should be able to use the `:pressed` variant selector here instead of the data attribute

design-toolkit/components/src/components/chip/styles.ts
```diff
@@ -64,11 +64,20 @@ export const DeletableChipStyles = tv({
...
+    ],
+    remove: [
+      'remove',
```
ArrayKnight, line 75: I think we may want to consider replacing the plain RAC Button here with the DTK Button so styles don't need to be reimplemented. Once that happens, the `has-` selectors above can target the `.group/button` className instead of having to add an arbitrary `remove` class as a selector target

design-toolkit/components/src/components/chip/styles.ts
seeddisperser: Interesting. `:pressed` didn't trigger the behavior correctly, which is why I used `[data-pressed]`.

colehart (approved):

PR #506, opened 2025-10-01
Feature: Map Toolkit Storybook

gromkii:
Closes [PATH-761](https://gohypergiant.atlassian.net/browse/PATH-761)

## ✅ Pull Request Checklist

## 📝 Test Instructions

- `pnpm --filter=@accelint/map-toolkit run preview` to start up storybook.
- Ensure storybook shows base map.

- [ ] Yes
- [x] No

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -815,21 +815,6 @@ importers:
       '@deckgl-fiber-renderer/shared':
         specifier: ^1.4.0
         version: 1.4.0(@deck.gl/core@9.1.14)(@deck.gl/mapbox@9.1.14(@deck.gl/core@9.1.14)(@luma.gl/core@9.1.10))(@types/react@19.1.15)(immer@10.1.3)(react@19.1.1)(use-sync-external-store@1.5.0(react@19.1.1))
-      '@luma.gl/constants':
-        specifier: ~9.1.9
-        version: 9.1.10
-      '@luma.gl/core':
-        specifier: ~9.1.9
-        version: 9.1.10
-      '@luma.gl/engine':
-        specifier: ~9.1.9
-        version: 9.1.10(@luma.gl/core@9.1.10)(@luma.gl/shadertools@9.1.10(@luma.gl/core@9.1.10))
-      '@luma.gl/shadertools':
-        specifier: ~9.1.9
-        version: 9.1.10(@luma.gl/core@9.1.10)
-      '@luma.gl/webgl':
-        specifier: ~9.1.9
-        version: 9.1.10(@luma.gl/core@9.1.10)
       maplibre-gl:
         specifier: ^5.7.1
         version: 5.8.0


packages/map-toolkit/src/deckgl/storybook-components/deckgl.tsx
```diff
@@ -0,0 +1,73 @@
...
+interface MapDecoratorProps {
+  width?: string | number;
+  height?: string | number;
+  backgroundColor?: string;
+  padding?: string | number;
+  center?: [number, number]; // [longitude, latitude]
+  zoom?: number;
+}
+
+export const MAP_STYLE =
```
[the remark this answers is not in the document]
Eshimusic: MAP_STYLE and PARAMETERS both, along with removing the related imports from `src/index.ts`. Then map-toolkit will be able to successfully build.

packages/map-toolkit/package.json
```diff
@@ -29,22 +29,52 @@
...
     "esbuild-fix-imports-plugin": "1.0.21",
+    "node-fetch": "2.6.7",
     "react": "^19.0.0",
+    "storybook": "^8.6.14",
     "tsup": "8.5.0",
     "vitest": "2.1.3",
     "vitest-broadcast-channel-mock": "0.1.0"
   },
   "dependencies": {
+    "@deck.gl/core": "9.1.14",
```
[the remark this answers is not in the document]
ArrayKnight: Those should be optional dependencies if they are associated with a specific feature. They should be dev dependencies as well. There shouldn't be any dependencies for this package

packages/map-toolkit/.storybook/main.ts
```diff
@@ -0,0 +1,36 @@
...
+ * It is needed in projects that use Yarn PnP or are set up within a monorepo.
+ */
+function getAbsolutePath(value: string): string {
+  return dirname(require.resolve(join(value, 'package.json')));
+}
+const config: StorybookConfig = {
+  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
+  addons: [
+    // getAbsolutePath('@storybook/addon-docs'),
+    getAbsolutePath('@storybook/addon-onboarding'),
```
switzerb, line 27: we probably don't need the `@storybook/addon-onboarding` add-on. you can just remove this...but! we *do* want to add as a dependency and include it here the `@storybook/addon-storysource` add-on.

packages/map-toolkit/src/deckgl/storybook-components/deckgl.tsx
```diff
@@ -0,0 +1,55 @@
...
+  zoom?: number;
+}
+
+/**
+ * A decorator that wraps stories in a map container with configurable properties
+ *
+ * @param options - Configuration options for the map container
+ * @returns A Storybook decorator function
+ */
+export const withDeckGL = (_options: MapDecoratorProps = {}): Decorator => {
```
switzerb, line 33: We are going to want the options to be able to initialize the Deck.gl layer with different params. To start, probably just `zoom` and `center` are enough.

Since we are using the `useMapLibre` hook here, you could pass-through those and use defaults for the rest. So something like, on the hook change the signature to:

`export function useMapLibre(deck: IControl | null, styleUrl: string, options: Partial<MapOptions>)` and then here in the decorator:

`useMapLibre(deckglInstance as IControl, MAP_STYLE, ...mapOptions)` and pass in the `center` and `zoom` properties from the `MapDecoratorProps`;

packages/map-toolkit/src/index.ts
```diff
@@ -16,3 +16,6 @@
 
 // biome-ignore-all assist/source/organizeImports: This comment is used to prevent the biome tool from altering the import statements in this file.
 
+export { INITIAL_VIEW_STATE, MAP_STYLE, PARAMETERS } from './deckgl/constants';
+export { useMapLibre } from './deckgl/hooks/use-maplibre';
```
switzerb, line 20: We don't actually want to export `useMapLibre` or `withDeckGL` from the library since they are just dev tools and not part of what we are shipping with the package. At the top of the `storybook-components/deckgl.tsx` file (and the useMapLibre file) you can add `// __private-exports` as the very first line and it will get skipped by the indexer.

packages/map-toolkit/.storybook/main.ts
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  framework: {
+    name: getAbsolutePath('@storybook/react-vite'),
+    options: {},
+  },
+  staticDirs: ['../public', '../assets'],
```
switzerb, line 33: You can remove line 33 since neither directory exists.

packages/map-toolkit/src/deckgl/storybook-components/deckgl.tsx
```diff
@@ -0,0 +1,55 @@
+/*
```
switzerb, line 1: We are going to need multiple decorators, so instead of putting both these into a `storybook-components` directory, it would be better to make a `decorators` folder with `deckgl` inside it. And then a `components` at the root with the `base-map.stories.tsx` file inside as a start. We can organize the components as we add stuff.

packages/map-toolkit/package.json
```diff
@@ -29,25 +29,70 @@
...
+    "@deck.gl/layers": "9.1.14",
+    "@deckgl-fiber-renderer/dom": "1.4.0",
+    "@deckgl-fiber-renderer/shared": "^1.4.0",
+    "@deck.gl/extensions": "^9.1.14",
+    "@deck.gl/geo-layers": "^9.1.14",
+    "@deck.gl/mapbox": "^9.1.14",
+    "@deck.gl/mesh-layers": "^9.1.14",
+    "@deck.gl/widgets": "^9.1.14",
+    "@luma.gl/core": "~9.1.9",
```
switzerb, line 88: You may not need all these `@luma.gl` dependencies -- I think they are transitive dependencies of `@deck.gl` itself so this might be duplicative inclusion. Not 100% sure of that.

packages/map-toolkit/README.md
```diff
@@ -30,3 +30,6 @@ In an effort to reduce the number of unnecessary dependencies being installed by
 
 ## Usage
 - Under Construction 🚧
+
+## Storybook Preview
+- To run storybook, use the command `pnpm preview` to run locally.
```
switzerb, line 35:
```suggestion
- To launch a preview locally, run at the command line `pnpm --filter=@accelint/map-toolkit preview`
```

packages/map-toolkit/.storybook/main.ts
gromkii: I'll need to take some time to read up on these addons, but updated to include these.

packages/map-toolkit/package.json
gromkii: Removing these for now, in that case.

packages/map-toolkit/src/deckgl/constants/index.ts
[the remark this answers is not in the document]
gromkii: Makes sense enough to me.

packages/map-toolkit/src/deckgl/hooks/use-maplibre.ts
```diff
@@ -0,0 +1,70 @@
...
+  doubleClickZoom: false,
+  dragRotate: false,
+  pitchWithRotate: false,
+  rollEnabled: false,
+};
+
+export function useMapLibre(
+  deck: IControl | null,
+  styleUrl: string,
+  _options: Partial<MapOptions>,
```
switzerb, line 32: why the underscore?

gromkii: Very good question. I feel that was added in by linting, which is odd.

PR #517, opened 2025-10-02
chore: create and use magic string enum for classification

kalisjoshua:
Relates to PATH-366

## ✅ Pull Request Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/.changeset/beige-pens-fix.md
+++ b/.changeset/beige-pens-fix.md
@@ -1,0 +1,5 @@
+---
+"@accelint/design-toolkit": patch
+---
+
+create and use magic strings enum for classifications
--- a/design-toolkit/components/src/tokens/primitive.json
+++ b/design-toolkit/components/src/tokens/primitive.json
@@ -141,7 +141,7 @@
     },
     "classification": {
       "missing": "#0b0b0b",
-      "unclass": "#007A33",
+      "unclassified": "#007A33",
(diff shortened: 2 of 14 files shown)


design-toolkit/components/src/tokens/primitive.json
```diff
@@ -141,7 +141,7 @@
     },
     "classification": {
       "missing": "#0b0b0b",
-      "unclass": "#007A33",
```
switzerb, line 144: Changing this name here counts as a breaking change, since it is published out as a token. Why rename it?

kalisjoshua: This seemed like an abbreviation for the sake of less typing but would result in "head-scratcher" problems because it is inconsistent with respect to name. In most situations there would be auto-complete that would be available to finish the string so I thought it would be better to be consistent. Open to changing it back but I thought this would be a better option.

.changeset/beige-pens-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/design-toolkit": patch
```
colehart: does the update from `unclass` > `unclassified` make this a "major" version bump per Brenna's comment about that being a breaking change?

Peechiz (approved):

PR #530, opened 2025-10-03
chore: address the console errors during test execution

kalisjoshua:
Address (provide fixes) for the console errors/warnings output during test execution.

## ✅ Pull Request Checklist

- [ ] Yes
- [x] No


design-toolkit/components/src/components/dialog/dialog.test.tsx
```diff
@@ -12,42 +12,25 @@
 
 import { render, screen } from '@testing-library/react';
 import { describe, expect, it } from 'vitest';
+import { Button } from '../button';
 import { Dialog } from './';
-import type { DialogTriggerProps } from './types';
-
-function setup({
```
ArrayKnight, line 18: Why did you get rid of the setup function? This is a pattern that I want our tests to follow

design-toolkit/components/src/components/drawer/index.tsx
```diff
@@ -70,28 +71,34 @@ export const DrawerEventHandlers = {
   toggle: (view: UniqueId) => bus.emit(DrawerEventTypes.toggle, { view }),
 } as const;
 
-function DrawerTrigger({ children, for: events }: DrawerTriggerProps) {
-  const { parent } = useContext(ViewStackContext);
+const DrawerTrigger = forwardRef<HTMLElement, DrawerTriggerProps>(
```
ArrayKnight, line 74: We shouldn't be using `forwardRef` with React 19. Ref is now just a prop

design-toolkit/components/src/components/text-area-field/text-area-field.test.tsx
```diff
@@ -52,7 +52,7 @@ describe('TextAreaField', () => {
     setup({
       size: 'small',
       description,
-      label: 'Label',
+      'aria-label': 'Label',
```
ArrayKnight, line 55: Rather than doing this, I'd like to take the approach that we did with DateField, where the label prop feeds into the aria-label prop: https://github.com/gohypergiant/standard-toolkit/blob/main/design-toolkit/components/src/components/date-field/index.tsx#L120

kalisjoshua: I think I've understood.

colehart (approved):

colehart:
RAC components throw console warnings about aria-label not being present on text fields /inputs if no label is present for that form field. This MR is going through all the tests - and in one case TextField itself - that didn't add a label or may not require a label always and adding that prop to avoid seeing those warnings when testing

design-toolkit/components/src/components/text-area-field/index.tsx
```diff
@@ -84,6 +84,7 @@ export function TextAreaField({ ref, ...props }: TextAreaFieldProps) {
   return (
     <TextField
       {...rest}
+      aria-label={labelProp}
```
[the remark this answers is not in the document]
kalisjoshua: I was only attempting to satisfy the console warnings to eliminate the "noise" in output. I would expect that RAC would understand that and not output warnings for things they don't expect to get fixed. Happy to look into this further if we think more effort should be put in; I was aiming for the lowest possible effort.

PR #522, opened 2025-10-02
feat: open-source deckgl-layer-symbol

joshuamora-programmed:
Open-source the deckgl-layer-symbol package which adds support for rendering MIL-STD-2525 symbologies as a Deck.gl layer. Adds storybook story to demonstrate usage.

Closes [PATH-762](https://gohypergiant.atlassian.net/browse/PATH-762)

## ✅ Pull Request Checklist

- [ ] Included link to corresponding [GitHub Issue](https://github.com/gohypergiant/standard-toolkit/issues).
- [x] The commit message follows [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) [extended](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) guidelines.
- [x] Added/updated unit tests and storybook for this change (for bug fixes / features).
- [x] Added/updated documentation (for bug fixes / features)
- [x] Filled out test instructions.
- [x] Added changeset (for bug fixes / features).

## 📝 Test Instructions

- `cd` to `packages/map-toolkit/deckgl-layer-symbol`
- `pnpm install`
- `pnpm run preview` to view storybook story

Supplemental test:
- Add `"@accelint/map-toolkit": "workspace:*"` as an entry to `package.json` within `apps/next`
- `pnpm install`
- Open a component file
- You should be able to import `SymbolLayer`:
`import { SymbolLayer } from '@accelint/map-toolkit/deckgl-layer-symbol';`
- You should be able to perform the following side-effect import to add `<symbolLayer />` as an intrinsic element:
`import '@accelint/map-toolkit/deckgl-layer-symbol/fiber';`

## ❓ Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## 💬 Other information

PATH-762-deckgl-layer-symbol-story

Screenshot 2025-10-02 at 11 46 45 AM

PR #537, opened 2025-10-06
feat: map instance draft

switzerb:
Closes 

## ✅ Pull Request Checklist

- [ ] Yes
- [ ] No


packages/map-toolkit/src/deckgl/hooks/use-map-events.ts
```diff
@@ -0,0 +1,33 @@
...
+import { bus } from '../components/mapClient';
+import { MapEvents } from '../components/mapClient/events';
+import type {
+  MapClickEvent,
+  MapHoverEvent,
+} from '../components/mapClient/types';
+
+// TODO: need to use the react library: https://github.com/gohypergiant/standard-toolkit/blob/main/packages/bus/README.md#react or manage the memory leak
+export function useMapEvents() {
+  const [hover, setHover] = useState<Partial<MapHoverEvent['payload']>>();
```
ArrayKnight, line 23: Instead of this being state, the hook should accept callbacks that get triggered by the event handlers

packages/map-toolkit/src/deckgl/hooks/use-map-events.ts
```diff
@@ -0,0 +1,33 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import { useState } from 'react';
+import { bus } from '../components/mapClient';
+import { MapEvents } from '../components/mapClient/events';
+import type {
+  MapClickEvent,
+  MapHoverEvent,
+} from '../components/mapClient/types';
```
ArrayKnight, line 19: Let's update the folder naming casing to be kebab case like everywhere else

packages/map-toolkit/src/deckgl/components/mapClient/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const handleMapHover = (pickingInfo: PickingInfo) => {
+    bus.emit(MapEvents.hover, {
+      layer: pickingInfo.layer,
+      coordinate: pickingInfo.coordinate,
+    });
+  };
+
+  // TODO: need a callback prop for onClick and onHover so that devs can pass in their own behavior
+  const handleMapClick = useCallback((pickingInfo: PickingInfo) => {
```
ArrayKnight, line 54: Let's allow `onClick` and `onHover` optional props to be passed in and called in these handlers as well

packages/map-toolkit/package.json
```diff
@@ -70,6 +72,7 @@
     "vitest-broadcast-channel-mock": "0.1.0"
   },
   "optionalDependencies": {
+    "@accelint/bus": "^1.3.0",
```
colehart, line 75: how do these interact with devDependencies? I'm noticing we're duplicating some but not all. shouldn't client-only also be an optional?

Peechiz (dismissed):
LGTM 🤷‍♂️

switzerb:
Needs a changelog - and after this merges a little restructuring: https://github.com/gohypergiant/standard-toolkit/pull/522 but it is looking solid. Going through in more detail now.

Eshimusic (approved):
LGMT too!

packages/map-toolkit/src/deckgl/base-map/base-map.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import { withDeckGL } from '../../decorators/deckgl';
+import { useMapEvents } from '../hooks/use-map-events';
+import type { Meta, StoryObj } from '@storybook/react';
+
+const meta: Meta = {
+  title: 'DeckGL',
+  decorators: [withDeckGL()],
```
switzerb, line 19: I think we should skip the decorator in this particular story, since the point is to demonstrate the use of the `BaseMap` so it will be clearer I think to just render it directly.

packages/map-toolkit/src/deckgl/base-map/base-map.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const BaseMap: Story = {
+  // Using the deckGL decorator, blank base map.
+  render: () => {
+    useMapEvents({
+      onHover: (payload) => console.log('hover', payload),
+      onClick: (payload) => console.log('click', payload),
+    });
+
+    // biome-ignore lint/complexity/noUselessFragments: testing
+    return <></>;
```
switzerb, line 37:
```suggestion
    return <BaseMap className='h-dvh w-dvw' />;
```

packages/map-toolkit/src/deckgl/hooks/use-map-events.ts
```diff
@@ -0,0 +1,35 @@
...
+  MapHoverEvent,
+  MapHoverPayload,
+} from '../base-map/types';
+
+type UseMapEventsProps = {
+  onHover?: (payload: MapHoverPayload) => void;
+  onClick?: (payload: MapClickPayload) => void;
+};
+
+export function useMapEvents({ onHover, onClick }: UseMapEventsProps) {
```
ArrayKnight, line 27: I think I want to get rid of this hook. Because it always sets up listeners, even if you're not using them. And if we expand on that, we'll just keep adding more listeners and that will multiply with each usage. I think we need to enforce the usage of `useOn` directly

packages/map-toolkit/src/index.css
```diff
@@ -0,0 +1,19 @@
+/*
```
ArrayKnight, line 1: Since we don't intend to export this as part of this package (right?) can we implement this just where we need it (I think just in Storybook)

PR #550, opened 2025-10-08
feat: add map mode

colehart:
Closes no ticket. Based on the mode chart in this excalidraw: https://app.excalidraw.com/l/1nHFfuetcLC/7HhttD3ReG3 

## ✅ Pull Request Checklist

## 📝 Test Instructions

Run storybook and click on the new map mode stories.
-  basic usage story shows the basics of changing the map mode between many different modes
-  multiple consumers story shows how multiple UI elements can subscribe to the modeChanged event to update themselves depending on the map mode
- auth flow story shows how you can use authorization if it's important that a given mode be retained until the feature that owns the mode releases it. You can choose to auto-accept mode change requests or require a more formal auth flow that a user can control. Use cases include (among many others) drawing mode - needing to not select a map entity accidentally, say, while drawing a shape via clicking on the map.

- [ ] Yes
- [x] No


packages/geo/src/index.ts
```diff
@@ -23,5 +23,8 @@ export { formatDegreesDecimalMinutes } from './coordinates/latlon/degrees-decima
 export { parseDegreesDecimalMinutes } from './coordinates/latlon/degrees-decimal-minutes/parser';
 export { formatDegreesMinutesSeconds } from './coordinates/latlon/degrees-minutes-seconds/formatter';
 export { parseDegreesMinutesSeconds } from './coordinates/latlon/degrees-minutes-seconds/parser';
+export { createFormatter } from './coordinates/latlon/internal/format';
+export type { FormatOptions } from './coordinates/latlon/internal/format';
+export { getOrdinal } from './coordinates/latlon/internal/ordinal';
```
colehart: unrelated to my changes - I just pulled in main and ran build

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -27,10 +27,79 @@ import type { IControl } from 'maplibre-gl';
...
+ *
+ * function MapListener() {
+ *   useOn(MapEvents.click, (data) => {
+ *     console.log('Map clicked:', data.payload.info);
+ *   });
+ *   return null;
+ * }
+ * ```
+ */
 export const bus = Broadcast.getInstance<MapEventType>();
```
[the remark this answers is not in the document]
colehart: removed

packages/map-toolkit/src/deckgl/base-map/map-mode-provider/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+        // Clear any pending authorization requests since mode changed successfully
+        pendingRequestsRef.current.clear();
+        return;
+      }
+
+      // Otherwise, send authorization request
+      // Generate authId and store the request for later lookup
+      const authId = uuid();
+
+      // Clear any existing pending requests and add the new one (only one pending request at a time)
```
ArrayKnight: You've got a mixed strategy going on. Right now this clear is only allowing a single request at a time. But you're using a map which implies support for multiple requests

As I intended it, you'd want to allow multiple requests, but only one per owner, which means storing it by owner id, not auth id. The auth id is just used to make sure that the decision aligns with an active pending request

Though I think you've touched on an interesting edge case as well, where we'd likely want to clear the map of requests when a decision has been made.

colehart: ah - good catch. I'll update

colehart: should pending requests be canceled if the owner changes the mode to a different one before dealing with a pending request?

colehart: answered in chat: one per requester, don't cancel pending requests if owner changes modes unless they change to default and the request is for default (same mode)

packages/map-toolkit/src/maplibre/hooks/use-maplibre.ts
```diff
@@ -13,20 +13,57 @@
...
+ */
 export function useMapLibre(
   deck: IControl | null,
   styleUrl: string,
   options: MapOptions,
 ) {
   const mapRef = useRef<MapLibre | null>(null);
+  // Using a ref for options to avoid re-creating the map when options object reference changes
+  // The map is only created once on mount, options changes after that are ignored
+  const optionsRef = useRef(options);
```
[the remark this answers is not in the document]
switzerb: This is an interesting architecture -- I've used an LRU cache in the financial domain to keep certain calculations that were referenced heavily in memory, to improve reporting performance, and it made a pretty large difference. You could put this information in a follow-up ticket and we could explore? I'd want to see some (even slight) evidence that we actually have a leak before implementing anything but it seems like a good idea to me if we do.

packages/map-toolkit/src/map-mode/store.ts
```diff
@@ -0,0 +1,466 @@
...
+      const allRequests = Array.from(this.pendingRequests.values());
+      this.pendingRequests.clear();
+
+      for (const request of allRequests) {
+        (
+          this.bus.emit as unknown as (
+            type: string,
+            payload: ModeChangeDecisionPayload,
+          ) => void
+        )(MapModeEvents.changeDecision, {
```
switzerb: I'm not sure I completely understand why the `this.bus.emit` requires casting to define the payload it accepts -- doesn't `emit` take a generic to be able to handle this? Is there a reason we can't use it here? This chunk of code is pretty hard to read and if it can be simplified (and other instances in this file, etc) that would certainly make it more readable.

packages/map-toolkit/src/map-mode/store.ts
```diff
@@ -0,0 +1,466 @@
...
+  private approveRequestAndRejectOthers(
+    approvedRequest: PendingRequest,
+    excludeAuthId: string,
+    decisionOwner: string,
+    reason: string,
+    emitApproval: boolean,
+  ): void {
+    // Collect all other pending requests to emit rejections for
+    const requestsToReject: PendingRequest[] = [];
+    for (const [, request] of this.pendingRequests.entries()) {
```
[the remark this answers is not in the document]
colehart: doh! good catch

packages/map-toolkit/src/map-mode/store.ts
colehart: update - I could have used `Broadcast.getInstance<MapModeEventType>()` and that would have solved the issue - I forgot to type it. Reverting to getInstance and properly typing it.

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -13,45 +13,144 @@
...
+  /**
+   * Unique identifier for this map instance (required).
+   *
+   * Used to isolate map mode state between multiple map instances (e.g., main map vs minimap).
+   * This should be a UUID generated using `uuid()` from `@accelint/core`.
+   *
+   * The same instanceId should be passed to `useMapMode()` when accessing map mode state
+   * from components rendered outside of the BaseMap's children (i.e., as siblings).
+   */
+  instanceId: UniqueId;
```
ArrayKnight: Can we make this just `id`, please?

packages/map-toolkit/src/map-mode/store.ts
```diff
@@ -0,0 +1,485 @@
...
+  /**
+   * Determine if a mode change request should be auto-accepted without authorization
+   */
+  private shouldAutoAcceptRequest(
+    desiredMode: string,
+    requestOwner: string,
+    currentModeOwner: string | undefined,
+    desiredModeOwner: string | undefined,
+  ): boolean {
+    // Owner returning to default mode
```
ArrayKnight: NIT (I bet other people would disagree with me): could be one return statement for the whole thing...

packages/map-toolkit/src/deckgl/base-map/provider.tsx
```diff
@@ -0,0 +1,173 @@
...
+    // Warn in development if instanceId changes (indicates a stability issue)
+    if (storeRef.current !== null && process.env.NODE_ENV === 'development') {
+      console.warn(
+        `[MapIdProvider] instanceId changed from "${storeRef.current}" to "${instanceId}". ` +
+          'This causes a memory leak as the old store remains in the registry. ' +
+          'Ensure instanceId is stable (create outside component or use useState).',
+      );
+    }
+
+    getOrCreateStore(instanceId);
```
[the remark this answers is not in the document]
colehart: We cannot. updating the implementation to a simple useEffect breaks it. I can simplify this however to get rid of the refs.
```
export function MapProvider({ children, id }: MapProviderProps) {
  // This breaks the application - see screenshot for demonstration
  // The issue: useEffect runs AFTER children render, so useMapMode() throws
  // "MapModeStore not found" error when children try to access the store

  useEffect(() => {
    getOrCreateStore(id);
    return () => {
      destroyStore(id);
    };
  }, [id]);

  return <MapContext.Provider value={id}>{children}</MapContext.Provider>;
}
```

Image

PR #566, opened 2025-10-15
fix[geo]: bundle commonjs dependencies for mgrs/utm

kalisjoshua:
Closes PATH-863

The problem was that a deeply nested dependency was not "safe" for use in modern browser contexts and needed to be patched in our build process so that it would work as expected and without error. The problem is documented in a `ROOT_CAUSE.md` file and additional options for more forward-looking fixes are included but until those can be implemented the fixes in this PR will be necessary.

Storybook was added as a tool for validating that the fix is indeed working; i.e. the end result is usable in a browser context without errors. 

## ✅ Pull Request Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/.changeset/real-otters-march.md
+++ b/.changeset/real-otters-march.md
@@ -1,0 +1,5 @@
+---
+"@accelint/geo": minor
+---
+
+Bundle commonjs dependencies to avoid browser incompatibility problems.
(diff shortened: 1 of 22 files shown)


ArrayKnight:
What's the thought process around the negative test? Is the intent to catch when the patching becomes unnecessary so we can remove it? My assumption is that it will never get updated (hasn't in 11 years...)

ArrayKnight:
I'd like to look at what other options we have. Considering the top level dependency hasn't been touched in 3 years, it's concerning to be relying on this stack at all. I'd like to get an evaluation of what other dependencies may be available or what we would have to do to carry over the minimal necessary functionality to our codebase

PR #571, opened 2025-10-20
fix: updates date-field state styles

seeddisperser:
Adds a pressed state to date-field and updates other states to match the designs


design-toolkit/components/src/components/date-field/styles.ts
```diff
@@ -35,6 +36,8 @@ export const DateFieldStyles = tv({
       'text-right',
       'placeholder-shown:fg-primary-muted',
       'focus-visible:fg-a11y-on-accent focus-visible:bg-accent-primary-bold focus-visible:outline-none',
+      'data-[pressed]:bg-transparent',
```
ArrayKnight: Don't use the data attribute version, only use the variant selector

switzerb:
We are closing this because it has been decided that the date component does not need a pressed state.

PR #589, opened 2025-10-29
Bug/full width fields

natalieyip:
Closes (https://gohypergiant.atlassian.net/browse/PATH-889)

## ✅ Pull Request Checklist

## 📝 Test Instructions

Look through all the following fields components in storybook and verify the correct width. 

1. ComboBoxField should behave like Select (min width, popover uses trigger width for min width, not 100% width)
2. DateField should fit content width (not 100% width, stay the same width according to the format)
3. SearchField should be 100% width
4. SelectField should have min width, not be 100%, and the popover should use the trigger width variable for its min width
5. TextAreaField should be 100% width
6. TextField should be 100% width
7. TimeField should behave like DateField (fit content format width, not 100% width)

- [ ] Yes
- [x] No

### DIFF
--- a/.changeset/funky-mails-lie.md
+++ b/.changeset/funky-mails-lie.md
@@ -1,0 +1,5 @@
+---
+"@accelint/map-toolkit": patch
+---
+
+Add map-mode/store getCurrentModeOwner method to access the current map mode owner.
--- a/packages/map-toolkit/src/map-mode/store.test.ts
+++ b/packages/map-toolkit/src/map-mode/store.test.ts
@@ -604,13 +604,15 @@ describe('MapModeStore', () => {
   });
 
   describe('Mode Ownership Tracking', () => {
-    it('tracks mode ownership', () => {
+    it('tracks mode ownership and allows owner to switch modes freely', () => {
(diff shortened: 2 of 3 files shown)


design-toolkit/components/src/components/input/index.tsx
```diff
@@ -116,6 +116,7 @@ export function Input({ ref, ...props }: InputProps) {
       data-size={size}
     >
       <div
+        data-testid='input-container'
```
ArrayKnight: You should not be adding testid data attributes for non-testing purposes

design-toolkit/components/src/components/text-field/styles.ts
```diff
@@ -14,7 +14,7 @@ import { tv } from '@/lib/utils';
 
 export const TextFieldStyles = tv({
   slots: {
-    field: 'group/text-field flex flex-col gap-xs',
+    field: 'group/text-field flex flex-col gap-xs [&_[data-testid=input-container]]:max-w-none',
```
ArrayKnight: The input className below targets the element you're attempting to style here

PR #596, opened 2025-10-31
Hotfix: use client directive inclusion

brandonjpierce:
Our regex had a `/gm` flag attached which was seemingly causing some issues with the `rgx.test()` call.


brandonjpierce:
After change:

Screenshot 2025-10-30 at 9 12 48 PM

PR #601, opened 2025-11-05
Feature: Flashcard

gromkii:
Closes [PATH-516](https://gohypergiant.atlassian.net/browse/PATH-516)

## ✅ Pull Request Checklist

## 📝 Test Instructions

- Run storybook preview locally or check PR build.
- Navigate to new `Flashcard` component tab.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
+++ b/packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
@@ -19,11 +19,8 @@ A quick visual identifier for multiple points of data for a given map entity.
     { label: 'key', value: 'value', id: 'unique-id-7' },
   ];
 
-  <Flashcard isLoading={isLoading}>
-    <FlashcardHero>
-      <FlashcardHeader>Identifier</FlashcardHeader>
-      <FlashcardSubheader>DATA</FlashcardSubheader>
-    </FlashcardHero>
+  <Flashcard isLoading={isLoading} header='Identifier' subheader='DATA'>
+    <FlashcardHero />
     <FlashcardAdditionalData>SECONDARY_DATA_01</FlashcardAdditionalData>
     <FlashcardAdditionalData>SECONDARY_DATA_02</FlashcardAdditionalData>
     <FlashcardDetailsList>
@@ -44,15 +41,11 @@ A quick visual identifier for multiple points of data for a given map entity.
 The Flashcard component accepts all standard HTML div attributes.
 
 - **`isLoading`** - Set visual state for loading on the Flashcard component.
+- **`header`** - Wrapper for Heading component, displays top level text for Flashcard.
+- **`isLoading`** - Wrapper for Text component, displays subheading text for Flashcard.
 
 ### FlashcardHero
 - Upper half of the Flashcard to provide identifier data for the object being observed.
-
-### FlashcardHeader
-- Main identifier for Flashcard.
-
-### FlashcardSubheader
-- Subheading for main identifier.
 
 ### FlashcardAdditionalData
 - Secondary data information that will appear below the initial header.
--- a/packages/design-toolkit/src/components/flashcard/flashcard.stories.tsx
+++ b/packages/design-toolkit/src/components/flashcard/flashcard.stories.tsx
@@ -18,9 +18,7 @@ import {
   FlashcardDetailsLabel,
   FlashcardDetailsList,
   FlashcardDetailsValue,
-  FlashcardHeader,
   FlashcardHero,
-  FlashcardSubheader,
 } from '.';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 
@@ -39,19 +37,18 @@ const meta = {
   component: Flashcard,
   args: {
     isLoading: false,
+    header: 'IDENTIFIER',
+    subheader: 'DATA',
   },
 } satisfies Meta<typeof Flashcard>;
 
 export default meta;
 type Story = StoryObj<typeof meta>;
 export const Default: Story = {
-  render: ({ children, ...args }) => {
+  render: (args) => {
     return (
       <Flashcard {...args}>
-        <FlashcardHero>
-          <FlashcardHeader>IDENTIFIER</FlashcardHeader>
-          <FlashcardSubheader>DATA</FlashcardSubheader>
-        </FlashcardHero>
+        <FlashcardHero />
         <FlashcardAdditionalData>SECONDARY_DATA_01</FlashcardAdditionalData>
         <FlashcardAdditionalData>SECONDARY_DATA_02</FlashcardAdditionalData>
         <FlashcardDetailsList data-testid='secondary'>
--- a/packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
+++ b/packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
@@ -21,9 +21,7 @@ import {
   FlashcardDetailsLabel,
   FlashcardDetailsList,
   FlashcardDetailsValue,
-  FlashcardHeader,
   FlashcardHero,
-  FlashcardSubheader,
 } from '.';
 
 function setup(isLoading = false) {
@@ -39,11 +37,8 @@ function setup(isLoading = false) {
   ];
 
   render(
-    <Flashcard isLoading={isLoading}>
-      <FlashcardHero data-testid='hero'>
-        <FlashcardHeader>IDENTIFIER</FlashcardHeader>
-        <FlashcardSubheader>DATA</FlashcardSubheader>
-      </FlashcardHero>
+    <Flashcard isLoading={isLoading} header='IDENTIFIER' subheader='DATA'>
+      <FlashcardHero data-testid='hero' />
       <FlashcardAdditionalData>SECONDARY_DATA_01</FlashcardAdditionalData>
       <FlashcardAdditionalData>SECONDARY_DATA_02</FlashcardAdditionalData>
       <FlashcardDetailsList data-testid='secondary'>
--- a/packages/design-toolkit/src/components/flashcard/index.tsx
+++ b/packages/design-toolkit/src/components/flashcard/index.tsx
@@ -14,6 +14,7 @@
 import { clsx } from '@accelint/design-foundation/lib/utils';
 import 'client-only';
 import { createContext, useContext } from 'react';
+import { Heading, Text } from 'react-aria-components';
 import { DetailsList } from '../details-list';
 import { DetailsListLabel } from '../details-list/label';
 import { DetailsListValue } from '../details-list/value';
@@ -27,21 +28,16 @@ import type {
 
 export const FlashcardContext = createContext<FlashcardProps>({
   isLoading: false,
+  header: '',
+  subheader: '',
 });
 
 /**
  * Example usage.
  *
  * ```tsx
- * <Flashcard isLoading={isLoading}>
- *  <FlashcardHero>
- *    <FlashcardHeader>
- *      {header}
- *    </FlashcardHeader>
- *    <FlashcardSubheader>
- *      {subHeader}
- *    </FlashcardSubheader>
- *  </FlashcardHero>
+ * <Flashcard isLoading={isLoading} header="Identifier" subheader="DATA">
+ *  <FlashcardHero />
  *  <FlashcardAdditionalData>
  *    {secondaryData}
  *  </FlashcardAdditionalData>
@@ -61,10 +57,10 @@ export const FlashcardContext = createContext<FlashcardProps>({
  * ```
  */
 export function Flashcard(props: FlashcardProps) {
-  const { isLoading, children, className, ...rest } = props;
+  const { isLoading, header, subheader, children, className, ...rest } = props;
 
   return (
-    <FlashcardContext.Provider value={{ isLoading }}>
+    <FlashcardContext.Provider value={{ isLoading, header, subheader }}>
       <div {...rest} className={clsx(styles.container, className)}>
         {children}
       </div>
@@ -75,7 +71,7 @@ Flashcard.displayName = 'Flashcard';
 
 export function FlashcardHero(props: FlashcardComponentProps) {
   const { children, className, ...rest } = props;
-  const { isLoading } = useContext(FlashcardContext);
+  const { isLoading, header, subheader } = useContext(FlashcardContext);
 
   if (isLoading) {
     return (
@@ -88,33 +84,17 @@ export function FlashcardHero(props: FlashcardComponentProps) {
 
   return (
     <div {...rest} className={clsx(styles.hero, className)}>
-      {children}
+      <FlashcardHeader className={styles.header}>{header}</FlashcardHeader>
+      <FlashcardSubheader className={styles.subheader}>
+        {subheader}
+      </FlashcardSubheader>
     </div>
   );
 }
 FlashcardHero.displayName = 'FlashcardHero';
 
-export function FlashcardHeader(props: FlashcardComponentProps) {
-  const { children, className, ...rest } = props;
-
-  return (
-    <div {...rest} className={clsx(styles.header, className)}>
-      {children}
-    </div>
-  );
-}
-FlashcardHeader.displayName = 'FlashcardHeader';
-
-export function FlashcardSubheader(props: FlashcardComponentProps) {
-  const { children, className, ...rest } = props;
-
-  return (
-    <div {...rest} className={clsx(styles.subheader, className)}>
-      {children}
-    </div>
-  );
-}
-FlashcardSubheader.displayName = 'FlashcardSubheader';
+const FlashcardHeader = Heading;
+const FlashcardSubheader = Text;
 
 export function FlashcardAdditionalData(props: FlashcardComponentProps) {
   const { children, className, ...rest } = props;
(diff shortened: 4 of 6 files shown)


.changeset/every-showers-strive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/design-toolkit": minor
+---
+
+Includes new Flashcard feature. A display component to quickly summarize information at a glance, with multiple options to show data.
```
switzerb:
```suggestion
Adds Flashcard, a display component to quickly summarize information at a glance, with multiple options to show data.
```
nit: this is the the world's most absurd nitpick...but "A display component..." was a sentence fragment and it makes my skin itchy.

packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
```diff
@@ -0,0 +1,70 @@
...
+    <FlashcardHero>
+      <FlashcardIdentifier>
+        <FlashcardHeader>Identifier</FlashcardHeader>
+        <FlashcardSubheader>DATA</FlashcardSubheader>
+      </FlashcardIdentifier>
+    </FlashcardHero>
+    <FlashcardSecondaryContainer>
+      <FlashcardSecondaryData>SECONDARY_DATA_01</FlashcardSecondaryData>
+      <FlashcardSecondaryData>SECONDARY_DATA_02</FlashcardSecondaryData>
+      <FlashcardDetailsContainer details={details}/>
```
switzerb: So, we chatted about this a bit already, but my inclination here is to name this `FlashcardDetails` and then have it compose in the `DetailsList` here in the story to demonstrate one example of usage -- rather than trying to pass in the data and have the component itself loop over the data. If `FlashcardDetails` is agnostic and serves as a relatively unopinionated container then I think that adds a lot of flexibility.

The flip side is that if the designers *want* to enforce opinionated control, then my suggestion here is a bad one, so we might want to confirm one way or the other.

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    return null;
+  }
+
+  return (
+    <h2 {...rest} className={secondaryData({ className })}>
+      {children}
+    </h2>
+  );
+}
+FlashcardSecondaryData.display = 'FlashcardSecondaryData';
```
switzerb:
```suggestion
FlashcardSecondaryData.displayName = 'FlashcardSecondaryData';
```

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  return (
+    <div {...rest} className={subHeader({ className })}>
+      {children}
+    </div>
+  );
+}
+FlashcardSubheader.displayName = 'FlashcardSubheader';
+
+export function FlashcardSecondaryContainer(props: FlashcardComponentProps) {
```
switzerb: I don't think you need the word `Container` on this component. `FlashcardSecondary` is descriptive enough I think.

packages/design-toolkit/src/components/flashcard/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  ComponentPropsWithRef<'div'>;
+
+export type FlashcardProps = FlashcardComponentProps & {
+  isLoading?: boolean;
+};
+
+// TODO: How strict do we need to be with the value types?
+export type FlashcardDetailProps = FlashcardMetaData & {
+  className?: string;
+};
```
colehart: this is defined but never used

packages/design-toolkit/src/components/flashcard/flashcard.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {
+  Flashcard,
+  FlashcardDetailsContainer,
+  FlashcardHeader,
+  FlashcardHero,
+  FlashcardIdentifier,
+  FlashcardSecondary,
+  FlashcardSecondaryData,
+  FlashcardSubheader,
+} from '.';
```
colehart: need a story to show FlashcardSecondaryDetails component. That's missing from examples.

packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+function setup(isLoading = false) {
+  const details = [
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
+    { label: 'key', value: 'value' },
```
colehart: nit: these could be made a little more unique for more realistic scenarios

packages/design-toolkit/src/components/flashcard/types.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { ComponentPropsWithRef, PropsWithChildren } from 'react';
+
+export type FlashcardComponentProps = PropsWithChildren &
+  ComponentPropsWithRef<'div'>;
+
+export type FlashcardProps = FlashcardComponentProps & {
+  isLoading?: boolean;
+};
+
+export type FlashcardDetailContainerProps = {
```
colehart: is there a reason not to include the `ComponentPropsWithRef<'div'>` type extension here too?


```suggestion
export type FlashcardDetailContainerProps = ComponentPropsWithRef<'div'> & {
  details?: FlashcardMetaData[];
  classNames?: { /* ... */ };
};
```

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const displayDetails = details.slice(0, 5);
+
+  return (
+    <div
+      {...rest}
+      className={detailsContainer({ className: classNames?.container })}
+    >
+      {/* Limit to the first 5 items in details array.*/}
+      {displayDetails.map((item) => {
+        const detailId = uuid();
```
colehart: I know my previous comment led to this, and this is a little funky to create new uuids for these on every render. It'll lead to unnecessary re-renders.

A more stable approach would be to require an id in the item data shape. The id should be created when the `details` data is created, not when it's rendered.

packages/design-toolkit/src/components/flashcard/types.ts
```diff
@@ -0,0 +1,35 @@
...
+    item?: string;
+    label?: string;
+    value?: string;
+  };
+};
+
+export type FlashcardMetaData = {
+  label: string;
+  value: string | number | null;
+};
```
colehart: re: my other comment about ids, add those here:


```suggestion
import type { UniqueId } from '@accelint/core'
...
export type FlashcardMetaData = {
  id: UniqueId;
  label: string;
  value: string | number | null;
};
```

colehart (approved):

packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
switzerb: Just to document the decision, design wants to ensure that the details remain static (not interactive) and we don't allow too much information into the flashcard, so the restriction here is a choice. We can revisit in the future if we need the capability to be broader.

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const { children, className, ...rest } = props;
+  const { isLoading } = useContext(FlashcardContext);
+
+  // While loading, don't display.
+  if (isLoading) {
+    return null;
+  }
+
+  return (
+    <h2 {...rest} className={secondaryData({ className })}>
```
colehart: doesn't make sense to have an h2 here when we're formatting the text style. swtich to a `<p>` or `<div>` tag

packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
```diff
@@ -0,0 +1,70 @@
...
+# Flashcard
+A quick visual identifier for multiple points of data for a given map entity.
+
+## Example Usage
+
+### Basic
+
+```tsx
+  const details = [
+    { label: 'key', value: 'value', id: unique-id-1 },
```
ArrayKnight: NIT: You should wrap those unique ids in quotes

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
```
ArrayKnight: Please move this comment below the imports

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  return (
+    <div {...rest} className={identifier({ className })}>
+      {children}
+    </div>
+  );
+}
+FlashcardIdentifier.displayName = 'FlashcardIdentifier';
+
+export function FlashcardHeader(props: FlashcardComponentProps) {
```
ArrayKnight: You can simplify this component to just be `export const FlashcardHeader = Heading`, then the subheader can be `export const FlashcardSubheader = Text`, and then implement the context provider in the hero to pass the classes down. This helps to eliminate wrapper components that do almost nothing

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  return (
+    <div {...rest} className={subHeader({ className })}>
+      {children}
+    </div>
+  );
+}
+FlashcardSubheader.displayName = 'FlashcardSubheader';
+
+export function FlashcardSecondary(props: FlashcardComponentProps) {
```
ArrayKnight: I think you can get rid of this component entirely. Since it seem to be solely responsible for a bit of layout, that can be handled by the Flashcard itself. I would potentially reach for a 5 column grid. The 1st, 3rd, and 5th columns being gutters, 2nd and 4th being named. Hero spans all columns, secondary data spans 2-4, and the details list is on the two named columns label and value

ArrayKnight: Scratch that, that's too complicated. Just apply the padding to the Flashcard and use negative margin on the hero and the details list

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  return (
+    <div {...rest} className={flashcardData({ className })}>
+      {children}
+    </div>
+  );
+}
+FlashcardAdditionalData.displayName = 'FlashcardAdditionalData';
+
+export const FlashcardDetailsList = (props: FlashcardDetailsListProps) => {
```
colehart: missing displayName for this one. Also, every other component uses `function` sytax versus `const`/arrow. Worth consistency of one pattern, so I'd suggest updating this to be `function`

packages/design-toolkit/src/components/flashcard/types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type FlashcardComponentProps = PropsWithChildren &
+  ComponentPropsWithRef<'div'>;
+
+export type FlashcardProps = FlashcardComponentProps & {
+  isLoading?: boolean;
+};
+
+export type FlashcardDetailsListProps = FlashcardComponentProps &
+  Partial<DetailsListProps>;
```
colehart: ?double checking my understanding: are there specific things you want from DetailsListProps or any overlapping types between these two types?

packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(screen.getByText('SECONDARY_DATA_02')).toBeDefined();
+  });
+
+  it('should only show 5 additional details', () => {
+    setup();
+
+    // First item to be hidden by selector nth-of-type
+    const detailEntries = document.getElementsByClassName(
+      'nth-of-type-[n+6]:hidden',
+    );
```
colehart: I think there's a much easier way to test this - you could instead look at what is actually rendered on the screen and test that the labels of the items past 5 .not.toBeInTheDocument()

generally speaking, you shouldn't be using document queries in your tests. Instead you should stick to the RTL queries off of screen.

For this one, since you're asserting something is not in the document, it should be `expect(screen.queryByText('STATE')).not.toBeInTheDocument()`

packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    // Should not render header text.
+    const header = screen.queryByText('IDENTIFIER');
+    expect(header).toBeNull();
+
+    const hero = screen.getByTestId('hero');
+    expect(hero).toBeDefined();
+
+    // Two skellington components
+    expect(hero.childElementCount).toEqual(2);
```
colehart: there should be some kind of grabbable prop on the skeleton components I think to assert they're specifically what the children are:
`screen.getByTestId()` is what I would grab, and I would add a data-testid attribute to the skeleton components if they dont' have any identifiable other attributes in the screen.debug() you could grab.

packages/design-toolkit/src/components/flashcard/flashcard.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    expect(secondaryContainer.childElementCount).toBe(5);
+  });
+
+  it('should not show secondary data field while loading', () => {
+    setup(true);
+    const header = screen.queryByText('IDENTIFIER');
+    expect(header).toBeNull();
+    const secondaryData = screen.queryByText('SECONDARY_DATA');
+    expect(secondaryData).toBeNull();
+  });
```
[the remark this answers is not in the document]
colehart: more tests could fill this out:
Test for FlashcardHeader rendering independently
Test for FlashcardSubheader rendering independently
Test that FlashcardDetailLabel/Value show skeletons when loading
Edge case: What happens if Flashcard children are used outside Flashcard context?

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+      <div {...rest} className={container({ className })}>
+        {children}
+      </div>
+    </FlashcardContext.Provider>
+  );
+}
+Flashcard.displayName = 'Flashcard';
+
+export function FlashcardHero(props: FlashcardComponentProps) {
+  const { children, className, ...rest } = props;
```
colehart: you're not actually using the className below. Should you be?

packages/design-toolkit/src/components/flashcard/types.ts
gromkii: This inclusion was due to a type conflict when using `{...rest}` in the `DetailsList` component. It allows us to accept any `DetailsList` props if needed, but they won't be required.

packages/design-toolkit/src/components/flashcard/flashcard.docs.mdx
```diff
@@ -0,0 +1,67 @@
...
+      <FlashcardHeader>Identifier</FlashcardHeader>
+      <FlashcardSubheader>DATA</FlashcardSubheader>
+    </FlashcardHero>
+    <FlashcardAdditionalData>SECONDARY_DATA_01</FlashcardAdditionalData>
+    <FlashcardAdditionalData>SECONDARY_DATA_02</FlashcardAdditionalData>
+    <FlashcardDetailsList>
+      {details.map((detail) => (
+        <Fragment key={detail.id}>
+          <FlashcardDetailsLabel>{detail.label}</FlashcardDetailsLabel>
+          <FlashcardDetailsValue>{detail.value}</FlashcardDetailsValue>
```
colehart: As exported, these are typos
```suggestion
          <FlashcardDetailLabel>{detail.label}</FlashcardDetailLabel>
          <FlashcardDetailValue>{detail.value}</FlashcardDetailValue>
```

Worth considering if you want DetailsList to be DetailList to keep it all singular or if you want to pluralize the Label and Value names

packages/design-toolkit/src/components/flashcard/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const { isLoading } = useContext(FlashcardContext);
+
+  if (isLoading) {
+    return (
+      <div {...rest} className={clsx(styles.hero, className, 'gap-s')}>
+        <Skeleton
+          className={clsx(styles.skeleton)}
```
colehart: no need for clsx when only one class present

```suggestion
          className={styles.skeleton}
```

colehart (dismissed):
Looks good!

joshuamora-programmed:
looks good

joshuamora-programmed (approved):

PR #603, opened 2025-11-05
feat: add coordinate-field component

jstdwt:
Closes https://gohypergiant.atlassian.net/browse/PATH-867

## ✅ Pull Request Checklist

## 📝 Test Instructions
Open storybook and test the coordinate field component

- [ ] Yes
- [x] No

### DIFF
--- a/packages/design-toolkit/src/components/coordinate-field/context.tsx
+++ b/packages/design-toolkit/src/components/coordinate-field/context.tsx
@@ -18,14 +18,55 @@ import type { ProviderProps } from '@/lib/types';
 import type { CoordinateFieldProps, CoordinateFieldState } from './types';
 
 /**
- * Context for CoordinateField component props
+ * CoordinateField Context Architecture
+ *
+ * This file defines two separate contexts following the React Aria component pattern:
+ *
+ * 1. CoordinateFieldContext (Props Context):
+ *    - Contains user-provided props for the component
+ *    - Used for React Aria's context-based prop merging
+ *    - Part of the public composition API
+ *
+ * 2. CoordinateFieldStateContext (State Context):
+ *    - Contains derived/computed runtime state
+ *    - Used to share state with child components (e.g., CoordinateSegment)
+ *    - Primarily for internal use
+
+ * Separation keeps the public API (props) distinct from internal
+ * implementation details (state), improving maintainability and composition.
+ */
+
+/**
+ * Props Context for CoordinateField component.
+ *
+ * Contains user-provided props (label, format, size, value, onChange, etc.)
+ * and is used by React Aria's useContextProps hook for context-based prop merging.
+ * This enables parent components to provide default props to nested CoordinateField
+ * components, supporting composition patterns.
+ *
+ * Part of the public API - external consumers can use this for component composition.
+ *
+ * @see CoordinateFieldStateContext for internal runtime state
  */
 export const CoordinateFieldContext =
   createContext<ContextValue<CoordinateFieldProps, HTMLDivElement>>(null);
 
 /**
- * Context for CoordinateField state
- * Allows child components (like CoordinateSegment) to access shared state
+ * State Context for CoordinateField component.
+ *
+ * Contains derived/computed runtime state (segmentValues, currentValue,
+ * validationErrors, registerTimeout, etc.) that is shared with child components
+ * like CoordinateSegment. This avoids prop drilling for deeply nested children.
+ *
+ * This follows the React Aria pattern of separating props context (public API)
+ * from state context (internal implementation). While exported for composition
+ * scenarios and testing, this is primarily for internal use.
+ *
+ * @see CoordinateFieldContext for user-provided props
+ * @example
+ * // Used internally by child components
+ * const state = useCoordinateFieldStateContext();
+ * const { segmentValues, isDisabled, registerTimeout } = state;
  */
 export const CoordinateFieldStateContext =
   createContext<CoordinateFieldState | null>(null);


packages/design-toolkit/src/components/coordinate-field/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+    isInvalid,
+    isRequired,
+    size,
+  };
+
+  // Memoize all coordinate format conversions to avoid redundant calculations
+  const allCoordinateFormats = useMemo(
+    () => getAllCoordinateFormats(state.currentValue),
+    [state.currentValue],
+  );
```
colehart: option: because this is only used within the popover component which is only opened sometimes, you could move this to generate when the popover is opened to avoid extraneous calculations:

```suggestion
const [allCoordinateFormats, setAllCoordinateFormats] = useState(null);
const handlePopoverOpen = useCallback(() => {
  setAllCoordinateFormats(getAllCoordinateFormats(state.currentValue));
}, [state.currentValue]);
```

packages/design-toolkit/src/components/coordinate-field/coordinate-utils.ts
```diff
@@ -0,0 +1,1056 @@
...
+        return null;
+    }
+
+    // Parse the formatted string to extract individual segment values
+    // This is the circular part: geo formatted it, now we parse it back apart
+    // Necessary because geo doesn't expose the components directly
+    const segments = parseCoordinateStringToSegments(coordString, format);
+    return segments;
+  } catch (error) {
+    console.error('[convertDDToDisplaySegments] Exception caught:', error);
```
colehart: curious: what's the reasoning for showing one error in prod and both errors in dev? Should we only show console.errors in non-prod? Could this be a place to use the @accelint/logger?

packages/design-toolkit/src/components/coordinate-field/coordinate-field.test.tsx
```diff
@@ -0,0 +1,1945 @@
...
+      if (copyButtons.length > 0) {
+        await user.click(copyButtons[0]);
+
+        await waitFor(() => {
+          expect(writeTextMock).toHaveBeenCalled();
+        });
+      }
+    });
+
+    it.skip('does not copy invalid coordinates', async () => {
```
colehart: you've got a few `.skip`s going on in this file. Recommend adding todos or fixing the failures?

colehart (commented):
Does it make sense to debounce the validation by 300-500ms versus running on every segment change when all segments are full? If so, would go in the useCoordinateFieldChange hook in the handleSegmentChange func

packages/design-toolkit/src/components/coordinate-field/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+    const paddingWidth = 2;
+
+    // Add format button width (icon button is roughly 2.5-3ch)
+    const buttonWidth = enableCopy ? 3.5 : 0;
+
+    // Add gap between input and button (gap-m = 1rem = ~1ch)
+    const inputButtonGap = enableCopy ? 1.5 : 0;
+
+    return `${segmentWidth + literalWidth + gapWidth + paddingWidth + buttonWidth + inputButtonGap}ch`;
+  }, [state.editableSegmentConfigs, state.segmentConfigs, enableCopy]);
```
colehart: This is a pretty beefy func with lots of "magic numbers" that could break if styles change. Consider pulling this out to a util

packages/design-toolkit/src/components/coordinate-field/coordinate-utils.ts
jstdwt: that seems like exactly what I need - thanks

packages/design-toolkit/src/components/coordinate-field/segment.tsx
```diff
@@ -0,0 +1,196 @@
...
+      }
+      aria-disabled={effectiveIsDisabled}
+      autoComplete='off'
+      inputMode={
+        allowedChars?.includes('0-9')
+          ? allowedChars.includes('\\-\\.')
+            ? 'decimal'
+            : 'numeric'
+          : 'text'
+      }
```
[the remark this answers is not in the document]
jstdwt: good question - I don't know if we do

packages/design-toolkit/src/components/coordinate-field/width-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+  }, 0);
+
+  // Calculate width of literal characters (colons, spaces, etc.)
+  const literalWidth = segmentConfigs
+    .filter((c) => c.type === 'literal')
+    .reduce((sum, c) => sum + (c.value?.length || 0), 0);
+
+  // Add gap spacing between segments
+  // Number of gaps = number of segments - 1
+  const gapCount = segmentConfigs.length - 1;
```
colehart: Need to guard against negative gap counts:

```suggestion
  // Number of gaps = number of segments - 1 (minimum 0)
  const gapCount = Math.max(0, segmentConfigs.length - 1);
```

packages/design-toolkit/src/components/coordinate-field/index.tsx
```diff
@@ -0,0 +1,463 @@
...
+            <Popover onOpenChange={handlePopoverOpenChange}>
+              <PopoverTrigger>
+                <Button
+                  variant='icon'
+                  size={size}
+                  color='mono-bold'
+                  className={classNames?.formatButton}
+                  aria-label='View coordinate in all formats'
+                  isDisabled={!copy.isFormatButtonEnabled}
+                  data-format-button
```
colehart: interesting: I'm more used to seeing `data-testid='format-button'` instead of custom data properties - at least in application code. I don't think it's an antipattern at all, just different than what I'm used to seeing. A search through the design-toolkit reveals that we're using data-size and other custom data- fields, so may just be my own unfamiliarity of the pattern in dtk.

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-field-state.ts
```diff
@@ -0,0 +1,253 @@
...
+    return segments;
+  });
+
+  const convertValueToSegmentsOrClear = useCallback(
+    (valueToConvert: CoordinateValue | null) => {
+      // Clear any pending validation when value changes
+      if (validationTimeoutRef.current) {
+        clearTimeout(validationTimeoutRef.current);
+        validationTimeoutRef.current = null;
+      }
```
colehart: you repeat this check in multiple places - could extract to a helper function and call:

```
const clearValidationTimeout = () => {
  if (validationTimeoutRef.current) {
    clearTimeout(validationTimeoutRef.current);
    validationTimeoutRef.current = null;
  }
};
```

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-field-state.ts
```diff
@@ -0,0 +1,253 @@
...
+      }, 400);
+
+      validationTimeoutRef.current = timeoutId;
+      registerTimeout?.(timeoutId);
+    } else {
+      // Clear any pending validation when segments become incomplete
+      if (validationTimeoutRef.current) {
+        clearTimeout(validationTimeoutRef.current);
+        validationTimeoutRef.current = null;
+      }
```
colehart: same

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-field-state.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+  };
+
+  const applyPastedCoordinate = (pastedValue: CoordinateValue) => {
+    // Clear any pending validation when applying pasted coordinate
+    if (validationTimeoutRef.current) {
+      clearTimeout(validationTimeoutRef.current);
+      validationTimeoutRef.current = null;
+    }
```
colehart: same

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-field.ts
```diff
@@ -0,0 +1,201 @@
...
+  const focus = useCoordinateFocus({
+    editableSegmentConfigs: state.editableSegmentConfigs,
+  });
+
+  const handlePasteValueApplied = (pastedValue: CoordinateValue) => {
+    state.applyPastedCoordinate(pastedValue);
+    registerTimeout(
+      setTimeout(() => {
+        focus.focusFirstSegment();
+      }, FOCUS_DELAY_MS),
```
colehart: if the timeout is 0, why have this at all?

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-copy.ts
```diff
@@ -0,0 +1,106 @@
...
+    textArea.value = text;
+    textArea.style.position = 'fixed';
+    textArea.style.left = '-999999px';
+    textArea.style.top = '-999999px';
+    document.body.appendChild(textArea);
+    textArea.focus();
+    textArea.select();
+    try {
+      document.execCommand('copy');
+      setCopiedFormat(null);
```
colehart: checking my understanding: should this instead set it to the `formatToCopy`?


```suggestion
      setCopiedFormat(formatToCopy);
```

switzerb (approved):
This looks fantastic.

PR #604, opened 2025-11-06
feat: map cursor

colehart:
[Closes PATH-904](https://gohypergiant.atlassian.net/browse/PATH-904)

## ✅ Pull Request Checklist

## 📝 Test Instructions

Spin up storybook and look at the new MapCursor stories.

This MR includes a small cleanup for viewport where the defaults weren't reflective of the possible undefined initial state.

- [ ] Yes
- [x] No


packages/map-toolkit/src/map-cursor/store.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    if (!hasPermission && modeOwner) {
+      // Reject the request - modeOwner is guaranteed to exist here
+      this.bus.emit(MapCursorEvents.rejected, {
+        rejectedCursor: cursor,
+        rejectedOwner: requestOwner,
+        currentOwner: modeOwner,
+        reason: `Cursor change to ${cursor} rejected: "${requestOwner}" is not the mode owner. To change cursor, request map mode change.`,
+        id: this.id,
+      });
```
Eshimusic: If a mode owner is active when a different owner makes a change cursor request, updating the cursor owner list won't affect the snapshot since mode owner always takes precedent. Also when the mode switches to default then the store might not have an updated cursor list immediately available if we reject all updates from non-mode owners. I think we should autoaccept all cursor requests will valid params.

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -26,7 +27,79 @@ import type { PickingInfo, ViewStateChangeParameters } from '@deck.gl/core';
...
+  | 'rootElement'
+  | 'target'
+  | 'changedPointers'
+  | 'pointers'
+>;
+/**
+ * Strips non-serializable properties from MjolnirPointerEvent for event bus transmission.
+ * Removes functions and DOM elements that cannot be cloned.
+ */
+function serializeMjolnirEvent(
```
joshuamora-programmed: TIL TS function overloading

packages/map-toolkit/src/map-cursor/map-cursor.stories.tsx
```diff
@@ -0,0 +1,1106 @@
...
+/**
+ * Demonstrates cursor priority with map mode ownership.
+ * Mode owners have priority for cursor changes.
+ */
+export const WithModeOwner: Story = {
+  render: () => {
+    const emitNotice = useEmit<NoticeQueueEvent>(NoticeEventTypes.queue);
+
+    // Listen for cursor rejection events
+    useOn<CursorRejectedEvent>(MapCursorEvents.rejected, (event) => {
```
joshuamora-programmed: not sure if this has been noticed with `useOn` before or if it's something specific to the map cursor functionality, but this event handler seems to be firing twice (leading to double notice emission):

map-cursor-use-on

for what it's worth, the function created by `getOrCreateRequestCursorChange` is only being fired once (that's what's printing "requesting cursor change for..." in the devtools console shown in the GIF)

packages/map-toolkit/src/map-cursor/map-cursor.stories.tsx
```diff
@@ -0,0 +1,1106 @@
...
+export const WithHover: Story = {
+  render: () => {
+    function SymbolHoverLayer() {
+      const { requestCursorChange, clearCursor } =
+        useMapCursor(BASIC_USAGE_MAP_ID);
+
+      const hoverCallback = useCallback(
+        (info: { picked: boolean }) => {
+          if (info.picked) {
+            requestCursorChange('pointer', 'symbol-layer');
```
joshuamora-programmed: question: should this be called if the cursor is already `pointer`? maybe this is something handled outside the scope of the callback function?

hover-callback-trigger

packages/map-toolkit/src/map-cursor/map-cursor.stories.tsx
```diff
@@ -0,0 +1,1123 @@
...
+      const { cursor, requestCursorChange, clearCursor } =
+        useMapCursor(BASIC_USAGE_MAP_ID);
+
+      // Guard to avoid unnecessary bus events on repeated hover callbacks
+      // note: the store dedupes repeated requests from the same owner for the same cursor
+      // but best practice is to not send unnecessary bus events in the first place
+      const hoverCallback = useCallback(
+        (info: { picked: boolean }) => {
+          if (info.picked) {
+            if (cursor !== 'pointer') {
```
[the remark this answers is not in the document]
colehart: Actually - changing the logic to this creates a weird feedback loop and a flaky UI when hovering symbols. With the current logic, when the symbol is being hovered, we have an info.picked, so we pass that first gate. We only want to chnage the cursor to a pointer it isn't already that. otherwise that first logic gate is a no-op - in a good way.

If we combine the two conditions, when we are hovering the symbol and the cursor _is_ a pointer already, we trigger the else, swapping it back to the default cursor - which we don't want.

So we want the current flow. We could change it around to avoid the nesting though. What do you think?
```
if (!info.picked) {
  clearCursor('symbol-layer');
} else if (cursor !== 'pointer') {
  requestCursorChange('pointer', 'symbol-layer');
}
```

packages/map-toolkit/src/map-cursor/map-cursor.docs.mdx
```diff
@@ -0,0 +1,751 @@
...
+const emitNotice = useEmit<NoticeQueueEvent>(NoticeEventTypes.queue);
+
+useOn(MapCursorEvents.rejected, (event) => {
+  if (event.payload.id === MAIN_MAP_ID) {
+    emitNotice({
+      message: event.payload.reason,
+      color: 'critical',
+    });
+  }
+});
```
switzerb: I can't believe I am saying this because I have spent the bulk of my career trying desperately to either write or get other people to write docs -- but this is a LOT to read in a "co-located with story" kind of format. Like, we've reached the level of needing a table of contents or something. It was super helpful to read and the very specific examples are wonderful...but there is now so much much content that I don't think it can be scanned easily.

Not requesting any changes here in the MR and don't want to dissuade from the enthusiasm but we might need some other solutions for "overflow" -- like making pages in Millipedia and linking to it so that it can be absorbed, with section links and a table of contents and so on.

packages/map-toolkit/src/map-cursor/use-map-cursor.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (!actualId) {
+    throw new Error(
+      'useMapCursor requires either an id parameter or to be used within a MapProvider',
+    );
+  }
+
+  // Subscribe to store using useSyncExternalStore with fan-out pattern
+  const cursor = useSyncExternalStore(
+    getOrCreateSubscription(actualId),
+    getOrCreateSnapshot(actualId),
```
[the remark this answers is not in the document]
colehart: I considered that, but when would a server component need a cursor? That is strictly a client-side concern. Right now, only the viewport includes support for RSCs because you could ostensibly have a viewport incoming from data. But I don't see where you would have a cursor coming in from data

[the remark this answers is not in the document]
switzerb: This came up as an error when I was pulling in the mapTK resources into the base template in both the viewport and the cursor coordinates library. i can try and track down the exact error I ran into but it made the libraries unusable.

packages/map-toolkit/src/map-cursor/map-cursor.docs.mdx
colehart: Can edit this down if you'd like - no bigs. What's the desired level of detail? Is there an example you'd like me to work off of to avoid verbosity in the future?

switzerb: No need to edit - the content is good and useful. I think the ask here is more "keep an eye on it as we move forward" and maybe "help me think of solutions". It's a great problem to have, it is just that we may have exceeded what the format is designed for. I really don't want to have to build, say, manual tables of contents at the top of the mdx files. The long term solution is something like https://fumadocs.dev/ ...but I was stalling on making something like this a project until we were a little further in our libraries and recipes.

packages/map-toolkit/src/map-cursor/use-map-cursor.ts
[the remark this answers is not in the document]
colehart: I'll fix this in all mapTK store components

francistse23:
tested through stories. lgtm. my approval probably doesn't carry much weight. will defer to Brenna and others to resolve comments

packages/map-toolkit/src/map-cursor/use-map-cursor.ts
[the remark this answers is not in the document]
colehart: done

PR #612, opened 2025-11-12
refactor: 💡 convert design toolkit to css modules

ArrayKnight:
Closes [PATH-1034](https://gohypergiant.atlassian.net/browse/PATH-1034)

For details on this PR and what is changing, read the docs: https://github.com/gohypergiant/standard-toolkit/pull/612/files#diff-04d73129f8f3cec639e1a5658558f43cb998610a1051beb6b637283eed1c93c4

## ✅ Pull Request Checklist

## 📝 Test Instructions

There are a lot of files in this PR and a lot of things to test -- any effort you can make in this direction would be useful:

- run `pnpm --filter=@accelint/design-toolkit run preview` to check each component in the library to make sure it looks and works correctly
- run `pnpm --filter '*apps/next' run dev` to ensure the example app runs correctly
- run `pnpm run build` to make sure the build works

- [x] Yes
- [ ] No

### DIFF
--- a/packages/design-toolkit/src/components/coordinate-field/coordinate-utils.ts
+++ b/packages/design-toolkit/src/components/coordinate-field/coordinate-utils.ts
@@ -583,10 +583,13 @@ export function convertDDToDisplaySegments(
     const segments = parseCoordinateStringToSegments(coordString, format);
     return segments;
   } catch (error) {
-    logger.withError(error).error('Failed to convert DD to display', {
-      value: String(value),
-      format: String(format),
-    } as any);
+    logger
+      .withContext({
+        value: String(value),
+        format: String(format),
+      })
+      .withError(error)
+      .error('Failed to convert DD to display');
     return null;
   }
 }
@@ -649,10 +652,13 @@ export function convertDisplaySegmentsToDD(
       lon: LON,
     };
   } catch (error) {
-    logger.withError(error).error('Failed to convert display to DD', {
-      segments: JSON.stringify(segments),
-      format: String(format),
-    } as any);
+    logger
+      .withContext({
+        segments: JSON.stringify(segments),
+        format: String(format),
+      })
+      .withError(error)
+      .error('Failed to convert display to DD');
     return null;
   }
 }
@@ -814,9 +820,12 @@ function convertToFormat(
     }
 
     // Log other errors in development
-    logger.withError(error).error(`Failed to convert to ${format}`, {
-      value: JSON.stringify(value),
-    } as any);
+    logger
+      .withContext({
+        value: JSON.stringify(value),
+      })
+      .withError(error)
+      .error(`Failed to convert to ${format}`);
     return {
       value: COORDINATE_ERROR_MESSAGES.CONVERSION_FAILED,
       isValid: false,
@@ -883,9 +892,12 @@ export function getAllCoordinateFormats(
 
     return result;
   } catch (error) {
-    logger.withError(error).error('Failed to get all coordinate formats', {
-      value: JSON.stringify(validValue),
-    } as any);
+    logger
+      .withContext({
+        value: JSON.stringify(validValue),
+      })
+      .withError(error)
+      .error('Failed to get all coordinate formats');
     return invalidResult;
   }
 }
@@ -1002,10 +1014,13 @@ export function parseCoordinatePaste(
       }
     } catch (error) {
       // Log parsing errors in development for debugging
-      logger.withError(error).warn(`Failed to parse as ${format}`, {
-        pastedText: pastedText.trim(),
-        format: String(format),
-      } as any);
+      logger
+        .withContext({
+          pastedText: pastedText.trim(),
+          format: String(format),
+        })
+        .withError(error)
+        .warn(`Failed to parse as ${format}`);
       // Continue trying other parsers
     }
   }
--- a/packages/design-toolkit/src/components/sidenav/sidenav.stories.tsx
+++ b/packages/design-toolkit/src/components/sidenav/sidenav.stories.tsx
@@ -110,6 +110,7 @@ export const Default: Story = {
                   <Heading>{section}</Heading>
                   {items.map((item) => (
                     <SidenavItem
+                      key={item.id}
                       id={item.id}
                       textValue={item.text}
                       isDisabled={item.disabled}


packages/design-toolkit/src/components/button/styles.module.css
```diff
@@ -0,0 +1,917 @@
...
+        @variant pressed {
+          @apply fg-critical-pressed bg-critical-pressed outline-2 outline-critical-pressed;
+        }
+      }
+
+      @variant disabled {
+        @apply fg-disabled outline-interactive-disabled;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ArrayKnight: True, I'm sort of holding off on optimization until things are stable. I think there's also a number of styles that come out of the ToggleButton styles, since it inherits Button styles

packages/design-toolkit/src/components/combobox-field/styles.module.css
```diff
@@ -0,0 +1,99 @@
...
+    @variant focus-within {
+      @apply outline-accent-primary-bold;
+    }
+
+    @variant pressed {
+      @apply outline-interactive-pressed;
+    }
+
+    @variant group-size-medium/combobox-field {
+      @apply min-w-[160px] max-w-[400px];
```
orteth01: wdyt about using the css props directly here (and line 48)
```suggestion
      min-width: 160px;
      max-width: 400px;
```

packages/design-toolkit/src/components/combobox-field/styles.module.css
```diff
@@ -0,0 +1,99 @@
...
+      @apply fg-disabled;
+    }
+  }
+
+  .error {
+    @apply fg-serious-bold text-body-xs;
+  }
+
+  .popover {
+    @apply -ml-s min-w-[calc(var(--trigger-width)+(var(--spacing-s)*2))];
```
orteth01: writing the css property is more characters in this case, but i think being able to add/use whitespace makes it a little more digestible
```suggestion
    @apply -ml-s;
    min-width: calc(var(--trigger-width) + (var(--spacing-s) * 2));
```

packages/design-toolkit/src/components/search-field/styles.module.css
```diff
@@ -0,0 +1,87 @@
...
+      @apply invisible;
+    }
+
+    @variant group-disabled/search-field {
+      @apply invisible;
+    }
+  }
+
+  .loading {
+    /* TODO: Not animation, TW animation CSS vars seem to be missing */
```
orteth01: i was going to say we _could_ inline it if `@apply` is acting funny
```
      animation: var(--animate-spin, spin 1s linear infinite);
```

but oddly enough, it's working for me locally

Screenshot 2025-11-17 at 11 35 53 AM

packages/design-toolkit/src/components/combobox-field/styles.module.css
ArrayKnight: Yea, that's what I've been doing, just missed these ones. Thanks for the pointer

packages/design-toolkit/src/components/input/styles.module.css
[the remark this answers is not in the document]
orteth01: > But the styles I think are working pretty nicely and are clearer to read

i agree. i like that things are grouped under the top level classes (vs the slots kind of being spread all over the file in the TV version).

packages/design-toolkit/src/documentation/upgrades.docs.mdx
```diff
@@ -0,0 +1,374 @@
...
+    }
+  }
+}
+```
+
+There are a few important things to notice above:
+
+* **@reference** - the use of the `@reference` directive at the top of your css module is required to avoid duplication of applied classes. [Details](https://tailwindcss.com/docs/functions-and-directives#reference-directive)
+
+* **@layer** -- assigns the classes in that group to a layer. In an application-specific component, that would most likely be `l3`.
```
orteth01: to match the other directives, we could include a docs link

```suggestion
* **@layer** -- assigns the classes in that group to a layer. In an application-specific component, that would most likely be `l3`. [Detaiils](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer)
```

and maybe we shoudl note that, generally, folks should stick to the already-defined layers, unless they have a pretty good reason to create their own...

brandonjpierce:
Maybe 🐛 when I change a control value for Badge it prompts me to save the change which actually creates a code change? The badge itself does not update.

Screenshot 2025-11-24 at 2 58 03 PM
Screenshot 2025-11-24 at 2 58 00 PM

brandonjpierce:
Possibly 🐛 the visual different for `Option` component size change is incredibly minor, almost as if a font size adjustment was accidentally removed

brandonjpierce:
🐛 focus styles absent from search field

Screenshot 2025-11-24 at 3 21 07 PM

switzerb:
> Maybe 🐛 when I change a control value for Badge it prompts me to save the change which actually creates a code change? The badge itself does not update.
> 
> Screenshot 2025-11-24 at 2 58 03 PM Screenshot 2025-11-24 at 2 58 00 PM

Whoa, that is interesting. I am not seeing this at all either locally or on https://design-toolkit.accelint.io/?path=/story/components-badge. I suspect that behavior is this: https://storybook.js.org/docs/essentials/controls#creating-and-editing-stories-from-controls -- maybe we should just disable it completely.

colehart:
> > Maybe 🐛 when I change a control value for Badge it prompts me to save the change which actually creates a code change? The badge itself does not update.
> > Screenshot 2025-11-24 at 2 58 03 PM Screenshot 2025-11-24 at 2 58 00 PM
> 
> Whoa, that is interesting. I am not seeing this at all either locally or on https://design-toolkit.accelint.io/?path=/story/components-badge. I suspect that behavior is this: https://storybook.js.org/docs/essentials/controls#creating-and-editing-stories-from-controls -- maybe we should just disable it completely.

I've noticed this before in the map-toolkit Storybook instance and find it really annoying. I'm for disabling it

packages/design-toolkit/src/components/coordinate-field/coordinate-field.stories.tsx
```diff
@@ -206,7 +206,7 @@ export const SmallSize: Story = {
  * The copy button can be hidden by setting showFormatButton to false.
  * This works for both medium and small sizes.
  */
-export const WithoutCopyButton: Story = {
```
joshuamora-programmed: is this "copy" -> "format" change intentional?

joshuamora-programmed:
it looks like there are spacing/color differences in the drawer (header bottom margin vs no bottom margin and drawer tab colors/padding)

css-modules-refactor--drawer-styles

joshuamora-programmed:
looks like there is a padding difference specifically on the bottom button on the sidenav

css-modules-refactor--sidenav-button-pad

joshuamora-programmed:
looks like slider region is missing following the changes

css-modules-refactor--slider-region

brandonjpierce (approved):

PR #623, opened 2025-11-19
feat!: event bus payload constraint

devindavies:
Closes [PATH-1024](https://gohypergiant.atlassian.net/browse/PATH-1024)

**Test Coverage and Documentation**

* Tests for the bus now explicitly verify serialization of various structured clone-compatible types, ensuring the new constraints are correctly enforced.
* Documentation and comments have been updated to clarify that payloads must be serializable by the structured clone algorithm.

**Dependency Updates**

* The `type-fest` package is added to provide the `StructuredCloneable` type.

**Downstream Type Updates**

* Types in `@accelint/design-toolkit` (such as `NoticeContent` and `NoticeDequeueEvent`) now require metadata to be `Record<string, StructuredCloneable>`, ensuring all event metadata is serializable.

## ✅ Pull Request Checklist

## 📝 Test Instructions

Attempt to provide the `emit` method with a payload that is not serializable, such as a function or a Symbol, and verify that a type error is emitted.

Screenshot 2025-11-18 at 7 05 01 PM

- [x] Yes
- [ ] No

This introduces a major breaking change to the `@accelint/bus` package by constraining the `Payload` type to only allow values that are serializable by the structured clone algorithm. This ensures that only safe, serializable data can be emitted through the bus, aligning the type system with the actual runtime behavior. 

This change will require all consumers of `@accelint/bus` to update their event payloads to comply with the structured clone algorithm, preventing runtime serialization errors and improving type safety.

### DIFF
--- a/.changeset/shy-hairs-cry.md
+++ b/.changeset/shy-hairs-cry.md
@@ -1,0 +1,5 @@
+---
+"@accelint/map-toolkit": patch
+---
+
+update picking info to remove more unserializable properties in map-toolkit
--- a/packages/map-toolkit/src/deckgl/base-map/base-map.docs.mdx
+++ b/packages/map-toolkit/src/deckgl/base-map/base-map.docs.mdx
@@ -284,8 +284,9 @@ The BaseMap component comes with the following defaults:
 
 ```typescript
 type MapClickPayload = {
-  info: NonFuncPickingInfo; // PickingInfo with viewport omitted for serialization
+  info: NonFuncPickingInfo; // Modified PickingInfo for serialization
   event: NonFuncMjolnirGestureEvent; // Event with functions omitted for serialization
+  id: UniqueId; // The map instance ID
 };
 ```
 
@@ -293,9 +294,23 @@ type MapClickPayload = {
 
 ```typescript
 type MapHoverPayload = {
-  info: NonFuncPickingInfo; // PickingInfo with viewport omitted for serialization
+  info: NonFuncPickingInfo; // Modified PickingInfo for serialization
   event: NonFuncMjolnirPointerEvent; // Event with functions omitted for serialization
+  id: UniqueId; // The map instance ID
 };
 ```
 
-Note: Function properties and the viewport are omitted from event payloads to allow proper serialization through the event bus.
+### `NonFuncPickingInfo`
+
+The picking info is modified for serialization through the event bus:
+
+**Omitted properties** (contain non-serializable functions/objects):
+- `viewport` - Contains methods and non-cloneable objects
+- `layer` - Contains layer methods and configuration
+- `sourceLayer` - Contains source layer methods and configuration
+
+**Added properties** (extracted before omission):
+- `layerId?: string` - The ID of the picked layer (extracted from `layer.id`)
+- `sourceLayerId?: string` - The ID of the source layer if applicable (extracted from `sourceLayer.id`)
+
+This transformation allows you to identify which layer was interacted with while maintaining compatibility with the structured clone algorithm required by the event bus.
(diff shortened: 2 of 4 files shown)


packages/bus/src/broadcast/types.ts
```diff
@@ -11,6 +11,7 @@
  */
 
 import type { UniqueId } from '@accelint/core';
+import type { StructuredCloneable } from 'type-fest';
```
colehart: would it be helpful to consumers to re-export this type? Unsure, but curious your thoughts

packages/bus/src/broadcast/index.ts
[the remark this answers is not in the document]
devindavies: I considered this, but the speed of these events getting through the bus is paramount, I'm not sure we would want to effectively double the execution time by running structuredClone just for runtime errors.  The type-guard should be enough for consumers, considering this is well documented by both us and MDN.

packages/bus/src/broadcast/types.ts
```diff
@@ -26,10 +27,12 @@ export type Listener<P extends { type: string; payload?: unknown } = Payload> =
     callback: (data: P) => void;
   };
 
+export type StructuredCloneableData = StructuredCloneable;
```
colehart: Why alias it? Could just reexport the typefest type directly, right?

packages/bus/src/broadcast/index.ts
```diff
@@ -283,7 +285,7 @@ export class Broadcast<
       return;
     }
 
-    if (!this.channel.onmessage) {
+    if (!this.listeners[type]?.length) {
       console.warn('No listeners registered for this event type:', type);
       return;
     }
```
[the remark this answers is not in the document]
colehart: If so, would be nice to add a comment or JSDoc explaining that. Something like: Note: If no listeners are registered locally for this event type, the message will not be broadcast to other contexts.

packages/bus/src/broadcast/index.test.ts
```diff
@@ -31,6 +32,93 @@ describe('broadcast', () => {
...
+
+  it('should warn if no listeners for event', () => {
+    const bus = Broadcast.getInstance<Payload<'test', string>>();
+    const consoleMock = vi.spyOn(console, 'warn');
+
+    bus.emit('test', 'test');
+
+    expect(consoleMock).toHaveBeenCalled();
+  });
+
```
colehart: would be good to add a test for non-broadcasts across channels if that's meant (q from my other comment:


```suggestion
it('should not postMessage when no local listeners are registered', () => {
    const bus = Broadcast.getInstance<Payload<'test', string>>();
    const consoleMock = vi.spyOn(console, 'warn');

    // No listeners registered locally - emit should warn and skip broadcasting
    bus.emit('test', 'message');

    expect(consoleMock).toHaveBeenCalledWith(
      'No listeners registered for this event type:',
      'test',
    );

    // postMessage is not called since no local listeners exist
    // @ts-expect-error Accessing protected property
    expect(bus.channel.postMessage).not.toHaveBeenCalled();
  });
```

packages/bus/src/broadcast/index.test.ts
```diff
@@ -196,4 +295,56 @@ describe('broadcast', () => {
...
+    [/ab+c/i],
+    [new Error('Test error')],
+    [
+      new Map([
+        ['key1', 'value1'],
+        ['key2', 'value2'],
+      ]),
+    ],
+    [new Set([1, 2, 3])],
+    [new Uint8Array([1, 2, 3, 4])],
```
colehart:
```suggestion
    [new Uint8Array([1, 2, 3, 4])],
    [{}],
    [{ nested: { deeply: { value: 42 } } }],
```

like this

packages/bus/src/broadcast/index.ts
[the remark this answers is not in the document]
devindavies: Hmm, that's a good point, I changed the if statement to more accurately match the warning, but this would only be valid if `target === 'self'`. I think this block should be removed.

colehart (dismissed):

colehart (approved):

PR #629, opened 2025-11-19
fix: add zoom param

colehart:
This MR adds in the zoom to try and fix issues seen with the current solution at large zoom scales.


colehart:
cleaned up and merged into viewport in a separate branch

PR #909, opened 2026-03-10
chore: try and fix deps

brandonjpierce:
## Overview

RAC is working towards simplifying their dependency chains a bit to consolidate into a single package. This will greatly simplify things on our end as currently we have to explicitly pull in a few `@react-aria/*` scoped packages that contain exports we need that are not included in `react-aria` base package.

This PR essentially swaps as many imports over to `react-aria` and `react-stately` as possible to simplify the dependency chain and lower the risk for version mismatches.

I also bumped RAC to latest version.

### Dep Value Strategy

- https://www.npmjs.com/package/react-aria?activeTab=code
- https://www.npmjs.com/package/react-aria-components?activeTab=code

Browse the `package.json` file in each of these tabs. This allows us to grab the exact version specified for a given release. 

### Semver Range

Given the historical lack of proper semver adherence in the past, I opted to maintain the `~` patch range in our defined versions vs. the `^` minor range. I removed the actual patch value from the defined value to simplify things (follows the pattern on some of our other dependencies).

### Overrides Simplification

Required block is now just:

```
"pnpm": {
  "overrides": {
    "@internationalized/date": "~3.12",
    "@react-aria/interactions": "~3.27",
    "@react-aria/utils": "~3.33",
    "@react-stately/utils": "~3.11",
    "@react-types/shared": "~3.33",
    "react-aria": "~3.47",
    "react-aria-components": "~1.16",
    "react-stately": "~3.45"
  }
},
```

Some of these are actually only added in as a "just in case" from a `pnpm run lint:rac` output. Notably `@internationalized/date` and `@react-types/shared`.

### Verification

Ran the usual gates of build, lint, test, index, format, etc..

Ran through map toolkit and design toolkit storybooks, everything looked kosher.


switzerb:
So as it turns out, the failing visual regression test is actually a hydration mismatch error that only happens with a 12 hour clock, when the `AM/PM` prints. What is happening is that there is a brief flash with the server rendering a literal space (and is in the visual snapshot)...but then the client rendering collapses the space and a second later it renders differently. You can see in the video here that is just me refreshing the page:

https://github.com/user-attachments/assets/a86ee4c7-57b7-4fb9-94ee-742051b42c88

So I suspect the failing test is a timing thing -- the original snapshot was taken with a "server" rendered 12 hour clock timefield but in this PR it is a "client" rendered 12 hour clock timefield.

I could push up this as a commit if you wanted, but to validate, I added this to line 120 of the `design-toolkit/src/components/time-field/index.tsx` component to render spaces without collapsing and to suppress the hydration warning and it fixes the issue. There is probably a more elegant way to manage this...

```
              {(segmentProp) => {
                // Render literal spaces consistently to avoid hydration mismatch
                if (segmentProp.type === 'literal' && segmentProp.text.trim() === '') {
                  return <span suppressHydrationWarning>{segmentProp.text}</span>;
                }
                return (
                  <DateSegment
                    segment={segmentProp}
                    className={composeRenderProps(
                      classNames?.segment,
                      (className) => clsx(styles.segment, className),
                    )}
                  />
                );
              }}
```

switzerb:
Closing this pull request since it is superseded by the release of react-aria-components 1.17 and a change to the way they manage dependencies: https://github.com/gohypergiant/standard-toolkit/pull/981

PR #940, opened 2026-03-19
feat: rubber band zoom

devindavies:
Closes [PATH-374](https://gohypergiant.atlassian.net/browse/PATH-374)

## ✅ Pull Request Checklist

- [ ] Yes
- [ ] No

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [ ] Implementation


packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -398,11 +403,49 @@ export function BaseMap({
...
+
+    function handleKeyUp(event: KeyboardEvent): void {
+      if (event.key === 'Shift') {
+        rbzRef.current?.disable();
+        mapRef.current?.getMap().dragPan.enable();
+      }
+    }
+
+    document.addEventListener('keydown', handleKeyDown);
+    document.addEventListener('keyup', handleKeyUp);
```
switzerb: I have some concerns about this useEffect and the global event listeners here. First, won't they apply to every map instance? so a keydown or keyup event on the map could potentially affect a minimap? And I am pretty sure that document level listeners fire even when, say, the map itself doesn't have focus which might lead to strange-ish side effects. Maybe at a minimum we can add a focus check here.

packages/map-toolkit/src/maplibre/rbz-handler.ts
```diff
@@ -0,0 +1,433 @@
...
+   * Called automatically by MapLibre during camera transitions, so avoid
+   * expensive DOM work here; changes are deferred via `requestAnimationFrame`.
+   */
+  reset(): void {
+    this._startPos = undefined;
+    this._rbzBounds = undefined;
+    this._isDrawing = false;
+
+    if (this._boxIsVisible) {
+      requestAnimationFrame(() => {
```
switzerb: This is potentially minor, but requestAnimationFrame returns an id and it might be worth tracking this in a property as "pending" or "active" so that you can run cancelAnimationFrame in the `destroy()` destructor. That way you won't have lingering callbacks if the RBZ is dropped mid-stream.

switzerb:
I like the simplicity here and overall it makes a lot of sense. In truth, I don't really understand what advantage this has over just enabling the boxZoom with MapLibre...but I also understand there is historical context here that I probably don't know about. Is having an origin at the center really that critical a feature?

But, all in all, the only thing that gives me pause is the useEffect and the listeners. Let me know, now that things have changed, if you want me to just take over this work and finish it up.

brandonjpierce:
> I like the simplicity here and overall it makes a lot of sense. In truth, I don't really understand what advantage this has over just enabling the boxZoom with MapLibre...but I also understand there is historical context here that I probably don't know about. Is having an origin at the center really that critical a feature?
> 
> But, all in all, the only thing that gives me pause is the useEffect and the listeners. Let me know, now that things have changed, if you want me to just take over this work and finish it up.

Primary differences:

- This solution retains aspect ratio of screen. Default maplibre RBZ is completely free form.
- This solution originates from center of mouse click. Default maplibre RBZ originates from top left corner.

This just matches current behavior in legacy systems that users are very familiar with.

colehart:
> > I like the simplicity here and overall it makes a lot of sense. In truth, I don't really understand what advantage this has over just enabling the boxZoom with MapLibre...but I also understand there is historical context here that I probably don't know about. Is having an origin at the center really that critical a feature?
> > But, all in all, the only thing that gives me pause is the useEffect and the listeners. Let me know, now that things have changed, if you want me to just take over this work and finish it up.
> 
> Primary differences:
> 
> * This solution retains aspect ratio of screen. Default maplibre RBZ is completely free form.
> * This solution originates from center of mouse click. Default maplibre RBZ originates from top left corner.
> 
> This just matches current behavior in legacy systems that users are very familiar with.

We could expose the relevant options to users instead of hardcoding those. Is that something you'd like me to implement? 

Screenshot 2026-04-13 at 2 51 59 PM

I'm not sure if we would want these options to be "set once on mount" or able to be dynamically changed after map load?

packages/map-toolkit/src/deckgl/base-map/index.test.tsx
```diff
@@ -21,26 +21,17 @@ vi.mock('../../maplibre/hooks/use-maplibre', () => ({
...
+    const { container } = render(
+      <BaseMap className='custom-map-class' id={uuid()} />,
+    );
 
-      const { container } = render(<BaseMap id={specificId} />);
-
-      // Verify component renders successfully with the provided id
-      expect(container.firstChild).toBeInTheDocument();
-    });
+    expect(container.firstChild).toHaveClass('custom-map-class');
```
colehart, line 35: just cleaning this up - I touched this file bc I thought about adding integration tests for RBZ but it required too much mocking.

switzerb (approved):

brandonjpierce:
> > > I like the simplicity here and overall it makes a lot of sense. In truth, I don't really understand what advantage this has over just enabling the boxZoom with MapLibre...but I also understand there is historical context here that I probably don't know about. Is having an origin at the center really that critical a feature?
> > > But, all in all, the only thing that gives me pause is the useEffect and the listeners. Let me know, now that things have changed, if you want me to just take over this work and finish it up.
> > 
> > 
> > Primary differences:
> > 
> > * This solution retains aspect ratio of screen. Default maplibre RBZ is completely free form.
> > * This solution originates from center of mouse click. Default maplibre RBZ originates from top left corner.
> > 
> > This just matches current behavior in legacy systems that users are very familiar with.
> 
> We could expose the relevant options to users instead of hardcoding those. Is that something you'd like me to implement?
> 
> Screenshot 2026-04-13 at 2 51 59 PM
> I'm not sure if we would want these options to be "set once on mount" or able to be dynamically changed after map load?

If we are using JERIC2O as a reference, it would need to be adjustable at runtime as there is currently a user setting that changes the behavior. As an aside "adjustable at runtime" just means the component should accept a prop that determines behavior, we shouldn't actually track anything in state per se ourselves.

colehart:
> > > > I like the simplicity here and overall it makes a lot of sense. In truth, I don't really understand what advantage this has over just enabling the boxZoom with MapLibre...but I also understand there is historical context here that I probably don't know about. Is having an origin at the center really that critical a feature?
> > > > But, all in all, the only thing that gives me pause is the useEffect and the listeners. Let me know, now that things have changed, if you want me to just take over this work and finish it up.
> > > 
> > > 
> > > Primary differences:
> > > 
> > > * This solution retains aspect ratio of screen. Default maplibre RBZ is completely free form.
> > > * This solution originates from center of mouse click. Default maplibre RBZ originates from top left corner.
> > > 
> > > This just matches current behavior in legacy systems that users are very familiar with.
> > 
> > 
> > We could expose the relevant options to users instead of hardcoding those. Is that something you'd like me to implement?
> > Screenshot 2026-04-13 at 2 51 59 PM
> > I'm not sure if we would want these options to be "set once on mount" or able to be dynamically changed after map load?
> 
> If we are using JERIC2O as a reference, it would need to be adjustable at runtime as there is currently a user setting that changes the behavior. As an aside "adjustable at runtime" just means the component should accept a prop that determines behavior, we shouldn't actually track anything in state per se ourselves.

I just hopped into Jeric2o prod to try and find that, and the only zoom-related user setting I'm seeing is zoom threshold. Also, RBZ is a double click and hold to drag, not a Shift + click drag

PR #974, opened 2026-04-13
fix: prevent unnecessary rerenders in coffin corner store

colehart:
@Peechiz reported a bug with useCoffinCorner that mirrored the one found in DrawShapeLayer where a useEffect was triggering a rerender during initial mount, which broke deckgl and MapLibre's connection such that when you update the viewport after initial render, the coffin corners show up in the marker's initial location, not the updated one.

The root cause was that layer registration (layerId, getEntityId) was stored in reactive state. When a second layer registered, the shared state reference changed, triggering useSyncExternalStore's post-commit snapshot check and scheduling a mid-mount re-render that desynced the viewports.

This work moves layer registration to a non-reactive module-level registry (plain Map writes invisible to React) and refactors the store to be multi-layer aware, holding only `selectedId` and `hoveredId` per layer. The store now updates exclusively from user interaction through the map bus.

## ✅ Pull Request Checklist

## 📝 Test Instructions

- Pull ngc2 and checkout `feat/replace-selection-store`
- `pnpm build` this branch in standard-toolkit
- file-link the locally built map-toolkit within the ngc2 `package.json` and `pnpm i`
- `pnpm dev` and hover an entity to see the proper behavior, pan the map using your arrow keys, see the proper behavior still.

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [x] Implementation

### DIFF
--- a/packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner.docs.mdx
+++ b/packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner.docs.mdx
@@ -41,12 +41,13 @@ The original approach used a separate icon atlas (sprite sheet) layer to draw co
 
 ## Architecture
 
-The system has three independent layers. The React hook is a convenience wrapper — you can use the store imperatively for non-React contexts.
+The system has three independent layers plus a non-reactive registry. The React hook is a convenience wrapper — you can use the store and registry imperatively for non-React contexts. Multiple layers per map are supported.
 
 | Layer | Module | Role |
 |-------|--------|------|
-| **React** | `useCoffinCorner(mapId, layerId)` | Reads state, sets `layerId` |
-| **Store** (framework-agnostic) | `coffinCornerStore` | Listens to map bus (click/hover events), emits domain events (selected/deselected), holds `{ selectedId, hoveredId }` |
+| **React** | `useCoffinCorner(mapId, layerId)` | Subscribes to the store, registers the layer's entity-ID accessor with the non-reactive registry, returns `selectedId` / `hoveredId` for the requested layer |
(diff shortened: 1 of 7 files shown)


Peechiz:
Confirming that this did fix the issue in NGC2

packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner.docs.mdx
```diff
@@ -233,42 +244,78 @@ function MyMap() {
...
 ```
 
+### Multiple layers on one map
+
+A single map can have any number of registered coffin-corner layers. Each layer maintains its own selection and hover state. Clicking empty space deselects all layers; clicking a new entity on a different layer does not affect other layers' selections.
+
+```tsx
+function MyMap({ mapId }) {
+  const { selectedId: klvSelected } = useCoffinCorner(mapId, 'klv-tracks');
+  const { selectedId: cotSelected } = useCoffinCorner(mapId, 'cot-symbols');
```
shadyendless, line 296: I don't think the `useCoffinCorner` hook should exist at all. It adds state duplication and confuses the ultimate goal of "you should be using the selection manager".

If people need to manage the state for their layers in a different way, then we should allow them to do that in the way that makes sense for their project. The `CoffinCornerExtension` only cares that the props are being passed into the layer that it is enabled on, and it handles everything else internally.

This feels like it is necessitating additional complexity in our end-user's codebase when that is not the intention, and coffin corners work perfectly fine without using this hook/store to begin with.

switzerb:
It would be useful to get a video of the bug and then a video of the fix to make it clearer what behavior needs to change.

switzerb:
> Confirming that this did fix the issue in NGC2

Are you using the useCoffinCorner hook in your code? How are you storing selection state?

colehart:
closing this MR in favor of the work in https://github.com/gohypergiant/standard-toolkit/pull/980 that replaces the problematic store and hook pattern with new props.

PR #976, opened 2026-04-14
feat: improve tree documentation and docs

switzerb:
Closes no ticket -- just a little clean up and documentation for a complex component. Since it's just docs and testing (no change to behavior) I am not adding a changeset.

## ✅ Pull Request Checklist

## 📝 Test Instructions

Should produce no changes in behavior. Just cleanup.

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [ ] Implementation


packages/design-toolkit/src/hooks/use-tree/actions/cache.ts
```diff
@@ -282,13 +473,52 @@ export class Cache<T> {
...
+   * Updates all nodes in the cache with shared properties.
+   *
+   * Applies partial updates to every cached node. Useful for bulk operations
+   * like clearing all selections or resetting visibility states.
+   *
+   * Note: `parentKey` and `children` properties are ignored (not applied to nodes).
+   *
+   * @param props - Partial node properties to apply to all nodes.
+   * @param props.parentKey - Ignored (not applied).
+   * @param props.children - Ignored (not applied).
```
colehart: nit: feels odd to call out these params when they're ignored / stripped out


```suggestion
   * Note: `parentKey` and `children` are stripped and not applied to nodes.
   *
   * @param props - Partial node properties to apply to all nodes.
```

packages/design-toolkit/src/hooks/use-tree/actions/cache.ts
```diff
@@ -282,13 +473,52 @@ export class Cache<T> {
...
+   *
+   * @example
+   * ```typescript
+   * // Update with TreeNode
+   * cache.setNode('node-1', { key: 'node-1', isSelected: true, children: [...] });
+   *
+   * // Update with CacheTreeNode (child keys)
+   * cache.setNode('node-1', { key: 'node-1', isSelected: true, children: ['child-1', 'child-2'] });
+   * ```
+   */
```
colehart: Correct me if I'm wrong here - the type only accepts TreeNode<T> for node, not CacheTreeNode. I think it should be updated to the below

```suggestion
   * Automatically converts TreeNode children to keys for internal cache storage.
   *
   * @param key - The unique key of the node to update.
   * @param node - The TreeNode data to store.
   *
   * @example
   * ```typescript
   * cache.setNode('node-1', { key: 'node-1', isSelected: true, children: [...] });
   * ```
```

packages/design-toolkit/src/hooks/use-tree/actions/cache.ts
```diff
@@ -237,13 +358,22 @@ export class Cache<T> {
   }
 
   /**
-   * Builds a tree structure from cache
+   * Rebuilds the tree structure from the flattened cache.
+   *
+   * Converts the internal flat lookup map back into hierarchical TreeNode objects.
+   * Optionally recomputes visibility state based on ancestor expansion before building.
    *
-   * TODO: optimization - rebuild only from change to root
```
colehart, line 242: Double checking - do we want to retain this TODO? or nah?

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -295,9 +409,146 @@ The `actions` object provides the following methods:
...
+  const applyChange = async (
+    updateFn: () => TreeNode[],
+    serverSync: (tree: TreeNode[]) => Promise<void>
+  ) => {
+    const previous = tree;
+    const updated = updateFn();
+
+    // Optimistic update
+    setTree(updated);
+    setHistory([...history, previous]);
```
colehart: this pattern has a stale closure bug in it. Since we're referencing the old value instead of a simple replace, we should recommend functional updaters. Prefer:

```suggestion
    setHistory((history) => [...history, previous]);
```

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -295,9 +409,146 @@ The `actions` object provides the following methods:
...
+
+    try {
+      await serverSync(updated);
+    } catch (error) {
+      // Rollback on error
+      setTree(previous);
+      setHistory(history.slice(0, -1));
```
colehart: same reasoning as above:

```suggestion
      setHistory((history) => history.slice(0, -1));
```

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -295,9 +409,146 @@ The `actions` object provides the following methods:
...
+          (tree) => api.updateTree(projectId, tree)
+        ),
+      // ... wrap other methods with optimistic updates
+    },
+    canUndo: history.length > 0,
+    undo: () => {
+      if (history.length > 0) {
+        setTree(history[history.length - 1]);
+        setHistory(history.slice(0, -1));
+      }
```
colehart: same issue as above here:

```suggestion
      // Read the latest history atomically so undo works even when
      // called in rapid succession; setTree is batched in the same tick.
      setHistory((currentHistory) => {
        if (currentHistory.length === 0) {
          return currentHistory;
        }
        
        setTree(currentHistory[currentHistory.length - 1]);
        return currentHistory.slice(0, -1);
      });
```

packages/design-toolkit/src/hooks/use-tree/actions/cascade.test.ts
```diff
@@ -339,4 +339,57 @@ describe('Cascade Selection', () => {
...
+      );
+      // Select child1 — parent1 becomes selected (its only selectable child)
+      result.current.onSelectionChange(new Set(['child1']));
+      // Move child1 into parent2 (which has unselected child2)
+      const updated = result.current.moveInto('parent2', new Set(['child1']));
+      const parent2 = updated.find((n) => n.key === 'parent2');
+      // parent2 now has child1 (selected) and child2 (unselected) → indeterminate
+      expect(parent2?.isIndeterminate).toBe(true);
+    });
+  });
```
colehart, line 386: nit - should have whitespace between `it` blocks

packages/design-toolkit/src/hooks/use-tree/actions/cascade.test.ts
```diff
@@ -339,4 +339,57 @@ describe('Cascade Selection', () => {
...
+      );
+      // Select child1 only — parent becomes indeterminate
+      result.current.onSelectionChange(new Set(['child1']));
+      // Move child1 to root level (before parent)
+      const updated = result.current.moveBefore('parent', new Set(['child1']));
+      const parent = updated.find((n) => n.key === 'parent');
+      // parent now only has child2 (unselected), no longer indeterminate
+      expect(parent?.isIndeterminate).toBe(false);
+      expect(parent?.isSelected).toBe(false);
+    });
```
colehart: non-blocking nit - you could DRY these two up via parameterization: 

```suggestion
    it.each(['moveAfter', 'moveBefore'] as const)(
      'should clear indeterminate state when selected child is moved out via %s',
      (method) => {
        const { result } = renderHook(() =>
          useTreeActions({ nodes: simpleTree, selectionCascade: true }),
        );
        // Select child1 only — parent becomes indeterminate
        result.current.onSelectionChange(new Set(['child1']));
        // Move child1 out from under parent (before or after — both remove it from parent's children)
        const updated = result.current[method]('parent', new Set(['child1']));
        const parent = updated.find((node) => node.key === 'parent');
        // parent now only has child2 (unselected), no longer indeterminate
        expect(parent?.isIndeterminate).toBe(false);
        expect(parent?.isSelected).toBe(false);
      },
    );
```

packages/design-toolkit/src/components/tree/index.tsx
```diff
@@ -204,13 +207,14 @@ export function Tree<T>({
...
-        }
+  const handleSelectionChange = useCallback(
+    (selection: Selection) => {
+      if (selection !== 'all') {
+        onSelectionChange?.(selection);
       }
-    : undefined;
+    },
+    [onSelectionChange],
+  );
```
colehart, line 217: these are functional changes that won't be picked up by consumers w/o a changeset. I don't think that's an issue in the current release since we have other changesets that will bump the version tho.

packages/design-toolkit/src/hooks/use-tree/actions/cache.ts
switzerb: Nope!

packages/design-toolkit/src/components/tree/index.tsx
switzerb: I don't think this should change behavior at all...but I suppose I could make this a patch version bump.

colehart: I don't have a big concern here at all since whenever this goes out, there will be other dtk stuff that would bump the version - just pointing it out

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -53,13 +53,20 @@ The `Tree` can be rendered with either hard-coded static data or with dynamic da
 
 ## Basic Example, dynamic data
 
-For dynamic data rendering, the Tree component accepts an `items` array and a render function as children. This pattern is more efficient for trees with many nodes or when the data comes from an external source.
+For dynamic data rendering, the Tree component accepts an `items` array and a render function as children. This pattern is more efficient for trees with many nodes or when the data comes from an external source. When you use a dynamic tree like this with external data, then you include the information about that node -- `isExpanded` or `isDisabled` as a property of that node. In this case, the Tree does NOT keep it's own state. It will render the data as passed in.
```
colehart: typo

```suggestion
For dynamic data rendering, the Tree component accepts an `items` array and a render function as children. This pattern is more efficient for trees with many nodes or when the data comes from an external source. When you use a dynamic tree like this with external data, then you include the information about that node -- `isExpanded` or `isDisabled` as a property of that node. In this case, the Tree does NOT keep its own state. It will render the data as passed in.
```

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -53,13 +53,20 @@ The `Tree` can be rendered with either hard-coded static data or with dynamic da
...
 
-```jsx
-const items = [
+```tsx
+import type { TreeNode } from '@accelint/design-toolkit/hooks/use-tree/types';
```
colehart: curious - do we want this import path to be deep when the others in this file are shallower?


```suggestion
import type { TreeNode } from '@accelint/design-toolkit';
```

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -22,10 +22,10 @@ It is composed of the following components:
 
 The Tree component supports two different collection patterns:
 
-1. **Static Collection**: Hard-coded tree structure using nested TreeItem components
+1. **Static Collection**: Hard-coded tree structure using nested TreeItem components.
 2. **Dynamic Collection**: Data-driven tree structure using the `items` prop and a render function
 
-You should use one approach or the other, not both. The component will throw an error if you try to use both approaches simultaneously.
+You should use one approach or the other, not both. The component will throw an error if you try to use both approaches simultaneously. The `items` object implies a dynamic collection, but the `expandedKeys` prop is trying to pass in controlled component data. In this example, you would remove the `expandedKeys` and put the `isExpanded` prop on the nodes in your `items` object that you want expanded.
```
colehart: the new sentence is describing stuff that occurs in the below example. More natural flow would be to put that after the example

packages/design-toolkit/src/components/tree/tree.docs.mdx
```diff
@@ -51,48 +74,143 @@ The `Tree` can be rendered with either hard-coded static data or with dynamic da
...
+</Tree>
+```
+
+The `items` object implies a dynamic collection, but the `expandedKeys` prop is trying to pass in controlled component data. In the above example, you would remove the `expandedKeys` and put the `isExpanded` prop on the nodes in your `items` object that you want expanded.
+
+
+When using dynamic data, let the component manage its state internally or use the state management hooks from the `use-tree` package.
+
+
+## Customizing Tree Item Appearance
```
colehart, line 96: extra whitespace

```suggestion
The `items` object implies a dynamic collection, but the `expandedKeys` prop is trying to pass in controlled component data. In the above example, you would remove the `expandedKeys` and put the `isExpanded` prop on the nodes in your `items` object that you want expanded.

When using dynamic data, let the component manage its state internally or use the state management hooks from the `use-tree` package.

## Customizing Tree Item Appearance
```

colehart (approved):

PR #978, opened 2026-04-14
Feat/light dark mode

switzerb:
Closes [93](https://gitlab.accelint.dev/core-ux/standard-toolkit/-/issues/93).

Adapts design foundation to utilize [light-dark](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/light-dark) syntax which is supported by modern browsers and will reduce our generated CSS bundle size by ~50%.

in our tokens, instead of having 

```
@theme static {
  --bg-surface-default: ;
 ...etc
}
@layer theme {
  :root {
    @variant dark {
       --bg-surface-default: var(--primitive-neutral-900, #151517);
       ...etc
     }
    @variant light {
      --bg-surface-default: var(--primitive-neutral-100, #eff1f2);
      ...etc
    }
  }
}
```

now we can output just a single token that understands both light and dark mode

```
@theme static {
  --bg-surface-default: light-dark(
    var(--primitive-neutral-100, #eff1f2),
    var(--primitive-neutral-900, #151517)
  );
...etc
```

## ✅ Pull Request Checklist

## 📝 Test Instructions

- Run the example app to ensure that it works correctly for applications: `pnpm --filter '*apps/next' run dev` 
- [Smoke test the colors in storybook](https://design-toolkit-git-feat-light-dark-mode-accelint.vercel.app/)

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [ ] Testing
- [ ] Implementation


apps/next/src/features/link/__screenshots__/link.visual.tsx/link-false-false-false-focus-dark-chromium-linux.png
switzerb, line 1: So, as it turns out, explicitly setting the color-scheme property creates a subtle change to the browser's native user agent focus color, enough to cause the VRT to fail on the focus states.

packages/design-toolkit/src/providers/theme-provider.tsx
```diff
@@ -77,6 +77,7 @@ export function ThemeProvider({
   useEffect(() => {
     if (document) {
       const { documentElement } = document;
+      documentElement.style.colorScheme = mode;
       documentElement.classList.remove('dark', 'light');
       documentElement.classList.add(mode);
```
colehart: curious if we could refactor things to not depending on .dark or .light at all? The point of this refactor is to get of classList deps like this right?

However, if we do this suggestion, it's a breaking change - so, up to you.
```suggestion
```

packages/design-toolkit/.storybook/preview.ts
```diff
@@ -32,10 +32,12 @@ const preview: Preview = {
...
         if (sbTheme === 'light') {
           theme = themes.light;
+          rootEl.style.colorScheme = 'light';
           rootEl?.classList.remove('dark');
           rootEl?.classList.add('light');
         } else {
           theme = themes.dark;
+          rootEl.style.colorScheme = 'dark';
           rootEl?.classList.remove('light');
           rootEl?.classList.add('dark');
```
colehart: if we did refactor the below, then this shortens to

```suggestion
          rootEl.style.colorScheme = 'light';
        } else {
          theme = themes.dark;
          rootEl.style.colorScheme = 'dark';
```

colehart: if you do this refactor, note that both the design-toolkit and map-toolkit storybook `assets/docs.css` files that update the css for the tables in .dark and .light modes should updated as well. You can replace everything in there with:

```css
@media (prefers-color-scheme: light) {
  .sbdocs-content table tbody tr:nth-of-type(odd) {
    background-color: rgb(0 0 0 / 0.04);
  }
}
@media (prefers-color-scheme: dark) {
  .sbdocs-content table tbody tr:nth-of-type(odd) {
    background-color: rgb(255 255 255 / 0.04);
  }
}
```

colehart (dismissed):
Lgtm - cool to see the spec driven approach here. One question / idea but non-blocking

packages/design-toolkit/.storybook/preview.ts
ArrayKnight: I personally would stick with the class and not implement the inline style. Since you should have the global style that applies the color scheme according to the class. But maybe there's a reason you needed to

packages/design-foundation/scripts/generate-tokens.mjs
```diff
@@ -132,30 +205,24 @@ function generateThemes(primitiveConfig, semanticConfig) {
...
-    /** Light theme **/
-    @variant light {
-${lightThemeTokens.map((line) => `      ${line}`).join('\n')}
-    }
+  const themeBlock = `@theme static {
+${lightDarkTokens.map((line) => `  ${line}`).join('\n')}
+}`;
+
+  const variantsBlock = `@layer theme {
+  @variant dark {
```
ArrayKnight: You're missing the * selector wrapping this style

ArrayKnight: Also because this is static I would put into a CSS file directly instead of a generation script

colehart:
Things look good in the example app re light mode and dark mode swapping. 

Low priority things I noticed unrelated to your PR:
- the dialog page messes up re: containing the overlay to its parent container / the nav bar is unusable to navigate to another story. You can change the link manually to go to another page, but that's not what we want, right?
- I don't understand at a glance what the second box on the Divider page is meant to be demonstrating - a visual gap vs a horizontal rule? Same goes for some of the other pages with multiple examples on them. Not sure if this is that important to beef up, but thought I'd mention it for DX improvement opportunites.

openspec/changes/refactor-light-dark-css/design.md
```diff
@@ -0,0 +1,124 @@
...
+documentElement.classList.add(mode);
+```
+
+To:
+```javascript
+documentElement.style.colorScheme = mode;
+```
+
+**Alternative considered:** Keep class-based approach and use `@media (prefers-color-scheme)` for `light-dark()`.
+- **Rejected because:** Would still require JavaScript to sync classes with color-scheme property, adding complexity.
```
colehart: This isn't happening in theme-provider. We're still using the classList approach alongside the `style.colorScheme = mode` approach. I'm not 100% sure how spec-driven development works with these docs, and it could be worth updating the spec to match our current decisions?

packages/design-toolkit/src/providers/theme-provider.tsx
```diff
@@ -77,6 +77,7 @@ export function ThemeProvider({
   useEffect(() => {
     if (document) {
```
colehart: curious - how would `document` ever be falsy in this check? `useEffect`s never run in SSR*, and document is never falsy in a browser. It's safe to remove this guard

*see the final caveat here: https://react.dev/reference/react/useEffect#caveats

openspec/changes/refactor-light-dark-css/design.md
```diff
@@ -0,0 +1,124 @@
...
+@layer theme {
+  :root {
+    --fg-primary: light-dark(var(--light-val), var(--dark-val));
+    /* ...all other tokens */
+  }
+  
+  @variant dark { color-scheme: dark; }
+  @variant light { color-scheme: light; }
+}
+```
```
colehart: Similar q re: not sure if the design docs need to be updated here - the output still uses `@theme static` and the shape of the `@variant`'s is a little different now to. Instead of the above it has
`@variant dark { * { color-scheme: dark; } }`

Again not sure if this needs to be updated, just noting the divergence

packages/design-foundation/src/index.css
```diff
@@ -62,6 +62,18 @@
...
+    @variant dark {
+      color-scheme: dark;
+    }
+
+    @variant light {
+      color-scheme: light;
+    }
+  }
+}
+
```
colehart, line 76: curious - why do we need the wildcard wrapper around the variants?

packages/design-foundation/scripts/generate-tokens.mjs
```diff
@@ -89,38 +89,111 @@ ${lines}
...
+}
+
+function generateLightDarkToken(
+  varName,
+  lightValue,
+  lightFallback,
+  darkValue,
+  darkFallback,
+) {
+  return `-${varName}: light-dark(var(${lightValue}, ${lightFallback}), var(${darkValue}, ${darkFallback}));`;
```
colehart, line 138: I'm not sure that we need the css fallbacks in here. Thinking through the import chain
- user imports `@accelint/design-foundation/styles`
- `./styles` resolves to `dist/index.css`, which import `@import './tokens/themes.css'`
- `themes.css` imports `@import './toekns.css'`, which loads primitives before any `var(--primitive-*)` reference resolves.

Consumers can't import `themes.css` directly, so I don't think we need them.

In the spirit of the effort to reduce bundle size, I ran both with and without the fallback here and here are the results: 

`themes.css` source (the generated artifact)
Metric | With fallbacks | Without | Delta
-- | -- | -- | --
Raw bytes | 13,077 | 9,752 | -3,325 (-25%)
Gzipped bytes | 1,821 | 1,059 | -762 (-42%)
Lines | 377 | 98 | -279 (-74%)

`apps/next` total CSS payload (the shipped artifact)
Metric | With fallbacks | Without | Delta
-- | -- | -- | --
Raw bytes | 258,170 | 249,566 | -8,604 (-3.3%)
Gzipped bytes | 18,083 | 17,672 | -411 (-2.3%)

So, less important for the consumer bundle if you really want to keep them. Just got curious and ran the numbers for decision-making.

openspec/changes/refactor-light-dark-css/design.md
switzerb: This is a good point -- the spec here is just wrong, actually, because tailwind requires the class based approach (or at least, it's the simplest way to work with tailwind) so I left them both in place rather than replacing. I'll update the spec.

packages/design-foundation/scripts/generate-tokens.mjs
switzerb: Yeah, it might be a level of defensiveness in coding that borders on the paranoid, but since the cost in bytes is low I'm inclined to just keep them in place.

packages/design-foundation/src/index.css
switzerb: https://tailwindcss.com/docs/functions-and-directives#variant-directive @variant is a conditional block requiring a selector -- if it isn't attached to a selector then the css doesn't get generated by tailwind, I believe. The  * ensures color-scheme is explicitly set on every element rather than relying solely on inheritance.

packages/design-toolkit/src/providers/theme-provider.tsx
switzerb: Interesting! The check was for SSR since document wouldn't be defined and even though this is a client component, I know that it still runs on the server. But I wasn't aware that the useEffect doesn't run at all and it is safe to assume you are in a browser...and although I have now done my homework on this topic and I see it isn't necessary, it still feels weird to remove it -- but I shall!

colehart (approved):

PR #980, opened 2026-04-16
feat!: add coffin corner multiselect support, remove hook and single id props

colehart:
Closes no issue. Updates the coffin corners extension to support multiselect and multihover state when multiselecting by box drag (see https://coreux-recipes.accelint.dev/?path=/story/entity-interaction-selection--drag-selection)

simplifies API by removing unnecessary store and hook and single hover/selected id props

## ✅ Pull Request Checklist

## 📝 Test Instructions

Run the new multiselect story under coffin corners when previewing map-toolkit. Note that we'll need to build a new recipe to see the multi hovered work in action.

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [x] Implementation


packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner.docs.mdx
```diff
@@ -350,6 +352,58 @@ Override the default blue (`#39B7FA`) bracket color:
 
 The alpha channel modulates bracket opacity. `[255, 0, 0, 127]` gives you semi-transparent red brackets.
 
+### Multiselect with `selectedEntityIds`
+
+For multiselect scenarios (e.g. with a SelectionManager), pass a `Set` instead of a single ID. The extension draws brackets on every entity in the Set.
+
+```tsx
+function MultiSelectMap({ mapId }) {
+  const [selected, setSelected] = useState<Set<number>>(() => new Set());
```
Peechiz: `number` should be `EntityId` ?

colehart: fixed both here and in the story

packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner-extension.ts
```diff
@@ -398,6 +398,49 @@ const DEFAULT_SELECTED_CORNER_FILL: Rgba255Tuple = [57, 183, 250, 255];
...
+ * contents of the map with the new Set on every change.
+ */
+function syncEntitySet(
+  entities: Map<EntityId, number>,
+  newIds: ReadonlySet<EntityId> | undefined,
+  oldIds: ReadonlySet<EntityId> | undefined,
+  attributeManager: { invalidate: (name: string) => void } | null,
+  attributeName: string,
+): void {
+  if (newIds !== oldIds) {
```
shadyendless: You cannot compare `Set`s like this, this will always be true, even if the underlying data is exactly the same. You would want to use a helper method from Radashi to do this comparison.

packages/map-toolkit/src/deckgl/extensions/coffin-corner/types.ts
```diff
@@ -20,10 +20,35 @@ import type { Rgba255Tuple } from '@accelint/predicates';
  * @template TLayerProps - The host layer's props type to intersect with.
  */
 export type CoffinCornerExtensionProps<TLayerProps = unknown> = {
-  /** The currently selected entity ID. */
+  /**
+   * The currently selected entity ID (single-select).
+   * Ignored when `selectedEntityIds` is provided.
+   */
   selectedEntityId?: EntityId;
```
shadyendless: We should get rid of the singular `selectedEntityId` and `hoveredEntityId` and only support the array versions in my opinion to simplify the interface.

Whether or not something can have multiple selections should not be determined by the layer, but rather the business logic surrounding the layer itself. The layer should always "support" multiple selected entities.

packages/map-toolkit/src/deckgl/extensions/coffin-corner/coffin-corner-extension.ts
```diff
@@ -543,32 +594,42 @@ export class CoffinCornerExtension extends LayerExtension {
...
+      );
+    } else {
+      syncEntityId(
+        this.state.selectedEntities,
+        params.props.selectedEntityId,
+        params.oldProps.selectedEntityId,
+        attributeManager,
+        'instanceSelectedEntity',
+      );
     }
```
shadyendless: We can get rid of the `else` statements on this and the hovered by removing the singular prop.

**NOTE**
If we need to support the singular prop for backwards compatibility, then we can but we should just wrap it in a set internal to this extension rather than have separate function calls so that all of the logic beyond the "interface" is the same.

colehart: I'm not sure that we do need to support it - this was written before the SelectionManager code was completed, and we didn't think then to test multilayer use cases. Multiselect was a future state for some reason I don't recall. I'm ok with updating to remove the single ids paths.

shadyendless (approved):

PR #981, opened 2026-04-17
chore: update react-aria-components

switzerb:
Closes [Issue 88](https://gitlab.accelint.dev/core-ux/standard-toolkit/-/work_items/88)

Recently, react-aria-components released a new version (1.17) that greatly simplifies the dependency management of it's ecosystem. As this has been a significant pain point, we want to take advantage, upgrade, and remove the careful (and brittle) overrides that we needed to have in place.

Although there are a large number of updates in this PR, most of it is just changing to subpath imports.

This should be a no-op PR in the sense that it should not be a breaking change, but it will require downstream applications to upgrade their `react-aria-component` dependencies and to add `react-aria` and `react-stately`.

## ✅ Pull Request Checklist

## 📝 Test Instructions

This should not change behavior, so this is all regression testing.

The usual suspects: 
- Run the example app to ensure that it works correctly for applications: pnpm --filter '*apps/next' run dev
- Take a look at the storybook run: https://design-toolkit-git-feat-upgrade-rac-accelint.vercel.app/

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [ ] Implementation


apps/next/package.json
```diff
@@ -43,11 +32,13 @@
     "postcss": "^8.5.6",
     "radashi": "^12.6.2",
     "react": "^19",
-    "react-aria-components": "~1.12.1",
+    "react-aria": "~3.48.0",
+    "react-aria-components": "~1.17.0",
```
switzerb, line 36: I left these at patch-level compatibility out of an abundance of caution -- let's say my trust with how RAC handles their dependencies is a little thin,

brandonjpierce:
`apps/nextjs/node_modules` prior state, before `pnpm i`

Screenshot 2026-04-24 at 1 30 54 PM

`apps/nextjs/node_modules` after running `pnpm i` (I did not delete the directory first

Screenshot 2026-04-24 at 1 35 59 PM

`apps/nextjs/node_modules` after removing all `node_modules` dirs and then running `pnpm i`

Screenshot 2026-04-24 at 1 40 06 PM

apps/next/package.json
```diff
@@ -19,20 +19,9 @@
...
-    "@react-aria/ssr": "~3.9.10",
-    "@react-aria/toast": "~3.0.7",
-    "@react-aria/utils": "~3.30.1",
-    "@react-stately/color": "~3.9.1",
-    "@react-stately/data": "~3.13.2",
-    "@react-stately/toast": "~3.1.2",
-    "@react-stately/utils": "~3.10.8",
-    "@react-types/overlays": "~3.9.0",
-    "@react-types/shared": "~3.32.0",
+    "@react-aria/interactions": "^3.25.5",
```
brandonjpierce, line 24: I think we are good to remove this line yeah? 

```
"@react-aria/interactions": "^3.25.5",
```

apps/next/package.json
brandonjpierce: Now that these deps are a bit easier to manage should we lessen the specificity across the board (much like we do with the deckgl deps) e.g.

```
"react-aria": "~3.48"
```

brandonjpierce: Also we shouldn't need to specify the `react-aria` dep directly unless we are importing it directly in the next app correct? Since in normal behavioral patterns it will be transiently installed with `react-aria-components`?

brandonjpierce: Granted some of this might be a side effect of our `peerDepencies` setup for `@accelint/design-toolkit` but mayyyyybe we can shift some of those to `dependencies` instead? A little tricky to say here. The `peerDependencies` makes sense for deps that you may also include in your project (like if you wanted to use `react-aria-components` to make a custom one off component).

brandonjpierce:
Confirming after a fresh install after removing `node_modules` (but not removing lockfile) no issues with multiple rac version! 



```
pnpm run lint:rac

> @accelint/developer-toolkit@0.0.0 lint:rac /Users/brandonpierce/Projects/standard-toolkit
> pnpm zx scripts/rac-deps.mjs

-------------------
| RAC DEPS REPORT |
-------------------
No version conflicts found! ✓
```

apps/next/package.json
switzerb: Unfortunately, no. The menu visual regression test requires the `setInteractionModality` function to manage focus state for the RAC components (or, at least, I couldn't find a workaround) --- and the react-aria wrapper package doesn't re-export the utility functions  like `setInteractionModality`, only the hooks. It *in* the `react-aria/useFocusVisible` exported, but not exported with that hook.

apps/next/package.json
switzerb: Yeah I see where you are going with this, since we don't actually need the `react-aria` library. Let me make sure it's not actually required to ship next to the `react-aria-components`...but if not, seems like taking it out of the peer dependency requirements is a smart move.

switzerb: Removing ` react-aria/react-stately` as peer dependencies produces an interesting bundling side-effect in the nextjs app -- `Error: dynamic usage of require is not supported
    at <unknown> (../../packages/design-toolkit/dist/_virtual/rolldown_runtime.js:15:48)` which seems like SSR choking on a `require` from the library.
    
Ensuring those libraries don't get bundled by marking them external in the tsdown config eliminates the error -- not sure that is any better than just leaving them as peer dependencies.

packages/design-toolkit/src/components/accordion/group.tsx
```diff
@@ -13,10 +13,11 @@
 
 import 'client-only';
 import { clsx } from '@accelint/design-foundation/lib/utils';
-import { composeRenderProps, DisclosureGroup } from 'react-aria-components';
+import { DisclosureGroup } from 'react-aria-components/DisclosureGroup';
 import { AccordionContext } from './context';
 import styles from './styles.module.css';
 import type { AccordionGroupProps } from './types';
+import { composeRenderProps } from 'react-aria-components/composeRenderProps';
```
hendrix-io, line 20: Are these import changes intentional? Is this a breaking change from RAC? That would explain why we must use the sub path exports.

hendrix-io (dismissed):
I saw 193 file changes and died inside but most of this seems like updating import paths so thats cool. I agree with your no-op statement in the description. Only other thing would be to confirm open comments from brandon about potentially dropping specifying the patch version in package.json.

packages/design-toolkit/src/components/accordion/group.tsx
switzerb: Yes, intentional, and no, not strictly required (or a breaking change) -- since they support subpaths and also imports from the root: https://react-aria.adobe.com/releases/v1-17-0#using-sub-paths. I chose to make this change since it just seemed like a good idea in terms of bundle optimization and I like how explicit it is, but it shouldn't change anything from the library usage side I don't think.

devindavies (dismissed):

apps/next/package.json
brandonjpierce: I'd like to investigate that a bit more to understand the root cause of why a rolldown runtime is being bundled into the application (e.g. missing `externals` config or similar)

brandonjpierce: As well still pursue the broadening of the semver ranges e.g. 
```
"react-aria": "~3.48"
```

brandonjpierce (approved):

PR #983, opened 2026-04-23
fix: draggable floating card title

devindavies:
This PR updates the styling in the FloatingCard header so that the title does not obscure the drag handle. 

## ✅ Pull Request Checklist

## 📝 Test Instructions

- Open the Floating Card in Storybook.
- Ensure the card is draggable over the title
- Check that the title does not overlap the header actions at any width, and truncates

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [ ] Implementation

## 💬 Other information

Before:

https://github.com/user-attachments/assets/67eac727-d529-433d-b34e-7fb807b02200

After:

https://github.com/user-attachments/assets/be1b3786-d3fb-4e50-a8cb-52ca2ed16545

### DIFF
--- a/packages/design-toolkit/src/components/floating-card/styles.module.css
+++ b/packages/design-toolkit/src/components/floating-card/styles.module.css
@@ -32,10 +32,14 @@
     @apply text-header-m pl-xs pointer-events-none h-full w-0;
   }
 
-  /** The header title is set to a min-width to prevent it from collapsing when the title is long, and truncating the overflow text.  */
-  /** Its width is based on the width of the floating card header via a container query, accounting for the padding and other elements in the header. */
+  /* Explicit width based on the header's container query width so long titles
+     truncate within the header rather than pushing siblings or overflowing.
+     The 10rem reserves space for the leading icon, header padding, gap, and
+     right-side action buttons; adjust when those change.
+     The @min-[1rem]/floating-card-header variant scopes the rule to cases
+     where the named container exists, so cqw never falls back to the viewport. */
   .headerTitle {
-    @apply h-full truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
+    @apply h-full truncate leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
   }
 
   .floatingCardProvider {
--- a/packages/design-toolkit/src/components/floating-card/styles.module.css
+++ b/packages/design-toolkit/src/components/floating-card/styles.module.css
@@ -22,11 +22,18 @@
   }
 
   .headerSide {
-    @apply gap-xs flex h-full items-center;
+    @apply gap-xs relative flex h-full items-center;
   }
 
+  .headerTitleContainer {
+    /** The header title container is set to a width of 0 to maximize the width of the drag handle, which we have no control over (dv-draggable).  */
+    @apply text-header-m pl-xs pointer-events-none h-full w-0;
+  }
+
+  /** The header title is set to a min-width to prevent it from collapsing when the title is long, and truncating the overflow text.  */
+  /** Its width is based on the width of the floating card header via a container query, accounting for the padding and other elements in the header. */
   .headerTitle {
-    @apply text-header-m pl-xs;
+    @apply h-full w-[calc(100cqw-10rem)] truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
   }
 
   .floatingCardProvider {
@@ -48,7 +55,7 @@
 
     :global(.dv-tabs-and-actions-container) {
       --dv-tabs-and-actions-container-height: 32px;
-      @apply p-xs text-header-m;
+      @apply p-xs text-header-m @container/floating-card-header;
     }
     :global(.dv-scrollable) {
       display: none;
--- a/packages/design-toolkit/src/components/floating-card/styles.module.css
+++ b/packages/design-toolkit/src/components/floating-card/styles.module.css
@@ -33,7 +33,7 @@
   /** The header title is set to a min-width to prevent it from collapsing when the title is long, and truncating the overflow text.  */
   /** Its width is based on the width of the floating card header via a container query, accounting for the padding and other elements in the header. */
   .headerTitle {
-    @apply h-full w-[calc(100cqw-10rem)] truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
+    @apply h-full truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
   }
 
   .floatingCardProvider {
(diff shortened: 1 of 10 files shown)


packages/design-toolkit/src/components/floating-card/styles.module.css
```diff
@@ -22,11 +22,18 @@
...
+  .headerTitleContainer {
+    /** The header title container is set to a width of 0 to maximize the width of the drag handle, which we have no control over (dv-draggable).  */
+    @apply text-header-m pl-xs pointer-events-none h-full w-0;
+  }
+
+  /** The header title is set to a min-width to prevent it from collapsing when the title is long, and truncating the overflow text.  */
+  /** Its width is based on the width of the floating card header via a container query, accounting for the padding and other elements in the header. */
   .headerTitle {
-    @apply text-header-m pl-xs;
+    @apply h-full w-[calc(100cqw-10rem)] truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
```
switzerb: Isn't the @min variant here redundant, since it is setting the width to the same thing that it already was?

packages/design-toolkit/src/components/floating-card/styles.module.css
```diff
@@ -48,7 +55,7 @@
 
     :global(.dv-tabs-and-actions-container) {
       --dv-tabs-and-actions-container-height: 32px;
-      @apply p-xs text-header-m;
+      @apply p-xs text-header-m @container/floating-card-header;
```
switzerb, line 64: Is the named container here because we can't inject our own classes into the dockview rendered DOM?

switzerb:
Totally unrelated to this PR, but since you are in here -- will you move the types into a separate `types.ts` file? Just to be consistent with the other components.

switzerb:
And just one more nit -- will you just throw a "before" and "after" screenshot into the PR description so there is a visual record of the bug and it's possible to see at a glance what the issue is?

packages/design-toolkit/src/components/floating-card/styles.module.css
devindavies: Actually, I should remove `w-[calc(100cqw-10rem)]` since:

1. It's redundant
2. If someone were to create a `@container` somewhere between this element and the named container it's deeply nested in, this style might conflict
3. the `@min` rule is more specific, and directly queries the named container.

packages/design-toolkit/src/components/floating-card/styles.module.css
devindavies: The container is here so that I can size the title based on this container, rather than the title's immediate parent, which has a width of 0. Otherwise, the title would either never be visible, or could not be truncated gracefully.

packages/design-toolkit/src/components/floating-card/types.ts
```diff
@@ -0,0 +1,207 @@
...
+     * Like other `initial*` props, this value is only read once at mount time.
+     * Changes after mount have no effect.
+     */
+    initialPinned?: readonly UniqueId[];
+    /** Additional CSS class names for styling */
+    className?: string;
+  }>
+>;
+
+export type HeaderAdapterOptions = {
```
colehart: why no jsdoc for this one when the others have it?

packages/design-toolkit/src/index.ts
```diff
@@ -264,12 +264,16 @@ export type {
...
 export type {
+  FloatingCardContextValue,
   FloatingCardHeaderAction,
+  FloatingCardHeaderProps,
+  FloatingCardProps,
   FloatingCardProviderProps,
-} from './components/floating-card/provider';
+  HeaderAdapterOptions,
+  MaybeFactory,
+} from './components/floating-card/types';
```
colehart, line 273: I think this exports unnecessary stuff re: comments in the types / uses:
- FloatingCardHeaderProps
- HeaderAdapterOptions
- MaybeFactory

colehart (commented):
you need to run build and push up the new export for the new types file

packages/design-toolkit/src/components/floating-card/styles.module.css
```diff
@@ -22,11 +22,18 @@
   }
 
   .headerSide {
-    @apply gap-xs flex h-full items-center;
+    @apply gap-xs relative flex h-full items-center;
```
colehart: I think this is an artifact of an earlier idea? not needed bc there's no `absolute` anywhere

packages/design-toolkit/src/components/floating-card/styles.module.css
```diff
@@ -22,11 +22,18 @@
...
   }
 
+  .headerTitleContainer {
+    /** The header title container is set to a width of 0 to maximize the width of the drag handle, which we have no control over (dv-draggable).  */
+    @apply text-header-m pl-xs pointer-events-none h-full w-0;
+  }
```
colehart, line 33: nit: could be more explanatory

```suggestion
    /* Zero width + pointer-events-none lets the title visually overflow the
     flex row without consuming drag-handle space or intercepting pointer
     gestures meant for dockview's drag handle (.dv-void-container). */
  .headerTitleContainer {
    @apply text-header-m pl-xs pointer-events-none h-full w-0;
  }
```

packages/design-toolkit/src/components/floating-card/styles.module.css
```diff
@@ -22,11 +22,18 @@
...
+
+  /** The header title is set to a min-width to prevent it from collapsing when the title is long, and truncating the overflow text.  */
+  /** Its width is based on the width of the floating card header via a container query, accounting for the padding and other elements in the header. */
   .headerTitle {
-    @apply text-header-m pl-xs;
+    @apply h-full truncate align-middle leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
```
colehart: this comment is a little misleading. The property being set isn't a min-width, it's a conditional width. Also I don't think you need `align-middle` in here. I tested out the following and it worked (but double check me on this if there were other use cases I'm not thinking of):

```suggestion
  /* Explicit width based on the header's container query width so long titles
     truncate within the header rather than pushing siblings or overflowing.
     The 10rem reserves space for the leading icon, header padding, gap, and
     right-side action buttons; adjust when those change.
     The @min-[1rem]/floating-card-header variant scopes the rule to cases
     where the named container exists, so cqw never falls back to the viewport. */
  .headerTitle {
    @apply h-full truncate leading-24 @min-[1rem]/floating-card-header:w-[calc(100cqw-10rem)];
  }```

devindavies: Yeah I think that's a fair change, I had a min-width rule in there before and forgot to update the comment when I took it out. Looks like you just dropped `align-middle` so as long as the title looks vertically centered it should be good.

packages/design-toolkit/src/components/floating-card/styles.module.css
devindavies: yes, removed

packages/design-toolkit/src/components/floating-card/types.ts
devindavies: overlooked, just added some

colehart (approved):

PR #989, opened 2026-05-05
fix: switch bug

sotomaque:
#### Bug: 
- Toggling a `<Switch>`  inside a scrollable container causes the container to auto-scroll, jumping content out from under the user.

#### Root cause: 
- Switch renders a `<input role="switch" type="checkbox"> ` that's visually hidden via `position: absolute; clip-path: inset(50%); width: 1px; height: 1px;` while the visible UI sits in sibling spans. 
- When the user clicks the visible label, react-aria's press handler calls `inputRef.current.focus()` to move keyboard focus onto the hidden input. 
- The browser's default behavior on `focus()` is to run `scrollIntoView({block: 'nearest'})` on the focused element — so it tries to bring a 1×1 pixel element into view, scrolling the nearest scrollable ancestor by however much the input's position differs from the viewport.

#### Fix:
- Apply `scroll-margin: 100vh` to the hidden input inside .switch, .checkbox, and .radio. This convinces `scrollIntoView({ block: 'nearest' })`  that the input is already visible, so the scroll is skipped.
- Visible inputs are unaffected — auto-scroll-on-focus was already a no-op for them.
- Scoped per component (no global selector), so the rule only touches the hidden a11y inputs these components render.

### DIFF
--- a/packages/design-toolkit/src/components/switch/styles.module.css
+++ b/packages/design-toolkit/src/components/switch/styles.module.css
@@ -27,14 +27,6 @@
     &.end {
       @apply flex-row;
     }
-
-    /* react-aria-components calls input.focus() without { preventScroll: true },
-       so toggling inside an overflow-y-auto ancestor scrolls to the 1px hidden
-       input. scroll-margin makes scrollIntoView({block:'nearest'}) treat it as
-       already visible. */
-    & input {
-      scroll-margin: 100vh;
-    }
   }
 
   .control {
--- a/packages/design-toolkit/src/components/switch/styles.module.css
+++ b/packages/design-toolkit/src/components/switch/styles.module.css
@@ -27,6 +27,14 @@
     &.end {
       @apply flex-row;
     }
+
+    /* react-aria-components calls input.focus() without { preventScroll: true },
+       so toggling inside an overflow-y-auto ancestor scrolls to the 1px hidden
+       input. scroll-margin makes scrollIntoView({block:'nearest'}) treat it as
+       already visible. */
+    & input {
+      scroll-margin: 100vh;
+    }
   }
 
   .control {
--- a/packages/design-toolkit/package.json
+++ b/packages/design-toolkit/package.json
@@ -431,6 +431,7 @@
     "./hooks/coordinate-field/use-timeout-cleanup": "./dist/hooks/coordinate-field/use-timeout-cleanup.js",
     "./hooks/kanban": "./dist/hooks/kanban/index.js",
     "./hooks/use-frame-delay": "./dist/hooks/use-frame-delay/index.js",
+    "./hooks/use-prevent-focus-scroll": "./dist/hooks/use-prevent-focus-scroll/index.js",
     "./hooks/use-tree/actions": "./dist/hooks/use-tree/actions/index.js",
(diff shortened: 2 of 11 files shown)


sotomaque:
https://github.com/user-attachments/assets/fddd56ef-d564-4c70-b897-e30ad1d92be4

PR with story that proves out bug here: https://github.com/gohypergiant/standard-toolkit/pull/991

switzerb:
> Screen.Recording.2026-05-05.at.6.39.52.PM.mov 
> PR with story that proves out bug here: #991

I think I am a little confused about the expectations around this fix. You *want* the container to scroll out of view with that jump? What is being fixed here? More directly -- on this branch with the style changes, I don't see any difference in the scroll behavior in the reproduction case. I feel like I am missing something. My apologies if I am being really dense...just trying to validate the fix.

sotomaque:
> > Screen.Recording.2026-05-05.at.6.39.52.PM.mov
> > PR with story that proves out bug here: #991
> 
> I think I am a little confused about the expectations around this fix. You _want_ the container to scroll out of view with that jump? What is being fixed here? More directly -- on this branch with the style changes, I don't see any difference in the scroll behavior in the reproduction case. I feel like I am missing something. My apologies if I am being really dense...just trying to validate the fix.

this video is not for this branch; this video is how the switch behaves today demonstrating the bug

switzerb:
> > > Screen.Recording.2026-05-05.at.6.39.52.PM.mov
> > > PR with story that proves out bug here: #991
> > 
> > 
> > I think I am a little confused about the expectations around this fix. You _want_ the container to scroll out of view with that jump? What is being fixed here? More directly -- on this branch with the style changes, I don't see any difference in the scroll behavior in the reproduction case. I feel like I am missing something. My apologies if I am being really dense...just trying to validate the fix.
> 
> this video is not for this branch; this video is how the switch behaves today demonstrating the bug

yes! that part makes sense -- but I am trying to say that running the same story on this branch does not appear to fix the bug and I feel like I am missing something. i.e. -- the style.module.css changes don't change behavior for me when I run it.

shadyendless:
Hey @sotomaque, could you update your PR description or add in a comment here the code that you wrote up for the repro story shown in the video that demonstrates the issue? Would like to paste it in locally and run it to verify the results.

EDIT: Disregard, just saw #991.

packages/design-toolkit/src/components/switch/styles.module.css
```diff
@@ -27,6 +27,14 @@
     &.end {
       @apply flex-row;
     }
+
+    /* react-aria-components calls input.focus() without { preventScroll: true },
+       so toggling inside an overflow-y-auto ancestor scrolls to the 1px hidden
+       input. scroll-margin makes scrollIntoView({block:'nearest'}) treat it as
+       already visible. */
+    & input {
+      scroll-margin: 100vh;
```
shadyendless: This does not fix the problem for switches. I did not check for the other components that this PR updates.

https://github.com/user-attachments/assets/325c692a-83b5-4f66-b7b9-0fd9a114c86c

switzerb:
Closing this PR since it doesn't address the issue, but it is still important reference for the bug fix. Documenting it here: https://gitlab.accelint.dev/core-ux/standard-toolkit/-/issues/118

PR #1007, opened 2026-05-20
feat: add type generics to useMapMode hook

Peechiz:
Closes [CORE-190](https://gohypergiant.atlassian.net/servicedesk/customer/portal/14/CORE-190)

## ✅ Pull Request Checklist

## 📝 Test Instructions

Use the hook and pass it a string union for Modes and Owners. Observe that this correctly types the output:

image
image
image

- [ ] Yes
- [x] No, still works without passing types.

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [ ] Implementation


colehart (commented):
I refactored the original approach to use T instead of User / follow generics naming patterns and update jsdocs and mdx. I also added defaults to prevent a breaking change (without defaults, TS will throw if not provided / it would be a breaking change).

packages/map-toolkit/src/map-mode/map-mode.docs.mdx
```diff
@@ -391,10 +394,22 @@ export function App() {
 
 ## React API Reference
 
-### `useMapMode(id?)`
+### `useMapMode<TMode, TOwner>(id?)`
```
switzerb: nit: I don't think we use the convention `TSomething` in the other parts of standard toolkit. There are places where it's just a single letter `T` for a generic -- or in this case you'd end up with something like `useMapMode<T,M>` so that's another option as well. But my preference would just be `useMapMode<Mode, Owner>` and skip the `T` prefix OR to just use the letter, not both.

Peechiz: I'd prefer the generics be named something useful, since putting them in the wrong order will mess up the type hint. I'm good with `<Mode, Owner>`.

colehart: Okie dokie

colehart (approved):

PR #1008, opened 2026-05-20
fix: set projection after MapLibre styles load

colehart:
Closes https://gitlab.accelint.dev/core-ux/standard-toolkit/-/issues/114

Maureen noticed an issue that arose after the PR to expose mapLibreOptions where, when setting the defaultView to 3D on the BaseMap, MapLibre would error with the following: 

image

This PR waits to set projection until after the style has been loaded to avoid the error.

## ✅ Pull Request Checklist

## 📝 Test Instructions
- to see the initial error, on `main` go to the next app BaseMap and set the `defaultView` to `3D`: 
Screenshot 2026-05-20 at 11 25 49 AM
 
-  notice the error on the Map page: 
Screenshot 2026-05-20 at 11 23 58 AM

- switch to this branch and rebuild to get the changes in map-toolkit
- run the next app and notice you are in globe view without the error: 

Screenshot 2026-05-20 at 11 25 27 AM

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [x] Implementation

### DIFF
--- a/packages/map-toolkit/src/deckgl/base-map/index.tsx
+++ b/packages/map-toolkit/src/deckgl/base-map/index.tsx
@@ -16,7 +16,7 @@ import { useEffectEvent, useEmit } from '@accelint/bus/react';
 import { Deckgl, useDeckgl } from '@deckgl-fiber-renderer/dom';
 import type { PickingInfo, ViewStateChangeParameters } from '@deck.gl/core';
 import 'client-only';
-import { useCallback, useId, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useId, useMemo, useRef } from 'react';
 import {
   Map as MapLibre,
   type MapRef,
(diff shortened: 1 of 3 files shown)


packages/map-toolkit/src/deckgl/base-map/index.tsx
brandonjpierce, line 512: Did you see this prop on the `MapLibre` component? https://visgl.github.io/react-map-gl/docs/api-reference/maplibre/map#projection

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -310,6 +310,11 @@ export function BaseMap({
   );
 
   // Spread order: default → consumer overrides → locked keys (last wins).
+  //
+  // Don't add MapLibre options here whose setter requires `isStyleLoaded()`
+  // (e.g. setProjection, setLight, setTerrain, setSky). react-map-gl applies
```
brandonjpierce: This is not true potentially? 

inside style.load -> 
https://github.com/visgl/react-map-gl/blob/master/modules/react-maplibre/src/maplibre/maplibre.ts#L323

https://github.com/visgl/react-map-gl/blob/master/modules/react-maplibre/src/maplibre/maplibre.ts#L461

this is also called in https://github.com/visgl/react-map-gl/blob/master/modules/react-maplibre/src/maplibre/maplibre.ts#L198 which should only be called, I believe, once a deckgl layer is added.

packages/map-toolkit/src/deckgl/base-map/index.tsx
colehart: That's the original (buggy) code - `projection` was being set above in the `mapOptions` to match the `cameraState.projection` value.

packages/map-toolkit/src/deckgl/base-map/index.tsx
colehart: updated to be more specific to projection issue.

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -324,21 +331,25 @@ export function BaseMap({
...
-      cameraState.view,
-      boxZoom,
-      filteredMapLibreOptions,
-    ],
+    [viewState, container, cameraState.view, boxZoom, filteredMapLibreOptions],
   );
 
+  // `setProjection` throws if called before the style is loaded. Initial
+  // application happens in `handleMapLoad`; this effect syncs later changes.
+  useEffect(() => {
```
brandonjpierce, line 343: I am still slightly curious as to why this isn't an issue in the baseline `react-map-gl` package 🤔 still leading me to believe we are chasing a potential red herring or something.

colehart: Am I remembering correctly that you want us to move away from react-map-gl? I was thinking it was due to some crud when in interleaved mode and we could refactor this out once we remove that package.

switzerb: I chatted with Devin about this, to ask if he remembered why he needed the library. The answer is...ambiguous but basically he said he tried to avoid using it and then ran into initialization issues.

Personally, I think we should move away from it as a library and refactor MapTK a bit, particularly the initialization / loading of all the interconnected systems. But that is a larger effort we don't have the cycles for right now, I think.

switzerb (approved):

PR #1009, opened 2026-05-21
Feat: Add camera pitchBy / rotateBy

Peechiz:
Closes n/a

## Description

In a client application, we implemented pitch/rotate via mouse right click drag events. Those event handlers only know the delta of the current drag gesture. In order to accurately set the state, you also need to know the current camera pitch/rotation. Getting that information into the drag handler from a react hook doesn't work, because the data is always stale.

A drag handler really wants to just send the delta and have the store add it to the value.  

## ✅ Pull Request Checklist

## 📝 Test Instructions

`useMapCamera` should now return `pitchBy` and `rotateBy` functions. Calling these with a delta should tilt & rotate the camera state by the given delta

- [ ] Yes
- [ ] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [x] Implementation


colehart (commented):
1. instead of `mod` prefix, let's follow the mapping/DOM ecosystem conventions for deltas with a suffix`By` so modPitch > `pitchBy`, modRotation > `rotateBy`

2. you should add rotation normalization of 0-360 and pitch clamping to MapLibre's range of 0-85 to avoid returning weird numbers
3. Add tests for this new request (clamping / normalization behavior), 

4. The MDX file for the camera needs to be updated to reflect these new functions

packages/map-toolkit/src/camera/types.ts
```diff
@@ -159,6 +159,22 @@ export type CameraSetPitchEvent = Payload<
   { id: UniqueId; pitch: number }
 >;
 
+/**
+ * Event type for modifying camera rotation angle by a delta.
+ */
```
colehart, line 165: worth noting with a `@remarks` tag in the JS doc that this is ignored in 3D view

packages/map-toolkit/src/camera/types.ts
```diff
@@ -159,6 +159,22 @@ export type CameraSetPitchEvent = Payload<
...
+export type CameraModRotationEvent = Payload<
+  typeof CameraEventTypes.modRotation,
+  { id: UniqueId; delta: number }
+>;
+
+/**
+ * Event type for modifying camera pitch angle by a delta.
+ */
```
colehart, line 174: worth noting with a @remarks tag in the JS doc that this is ignored outside of 2.5D view

packages/map-toolkit/src/camera/types.ts
Peechiz: added

packages/map-toolkit/src/camera/types.ts
Peechiz: added

packages/map-toolkit/src/camera/store.ts
```diff
@@ -228,12 +240,18 @@ const DEFAULT_CAMERA_STATE: CameraState = {
...
       const currentState = get();
       // Use buildCameraState to ensure proper discriminated union type
       replace(buildCameraState({ ...currentState, ...updates }));
     },
+    pitchBy: (delta: number) => {
+      cameraBus.emit(CameraEventTypes.pitchBy, { id: mapId, delta });
+    },
+    rotateBy: (delta: number) => {
+      cameraBus.emit(CameraEventTypes.rotateBy, { id: mapId, delta });
+    },
```
devindavies, line 262: This is a new pattern for this store, can you add some reasoning to the MR description for this new pattern? I'm not sure the store should be responsible for emitting these bus events

packages/map-toolkit/src/camera/types.ts
colehart: incorrect - this is ignored outside of 2.5D view, not in 3D view

.changeset/full-dancers-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/map-toolkit": minor
+---
+
+Adds camera modPitch and modRotation events, allowing to mutate pitch and rotation by delta rather than setting directly.
```
colehart: this needs to be updated after your changes.

packages/map-toolkit/src/camera/store.ts
```diff
@@ -397,6 +415,40 @@ export const cameraStore = createMapStore<CameraState, CameraActions>({
       },
     );
 
+    const unsubRotateBy = cameraBus.on(
+      CameraEventTypes.rotateBy,
+      ({ payload }) => {
+        if (payload.id !== mapId) {
```
colehart: would be good to defensively catch NaNs here. Note: worth also adding a test to catch this as well.

```suggestion
        if (payload.id !== mapId || !Number.isFinite(payload.delta) {
```

packages/map-toolkit/src/camera/store.ts
```diff
@@ -397,6 +415,40 @@ export const cameraStore = createMapStore<CameraState, CameraActions>({
...
+          const newRotation = bearingToAzimuth(state.rotation + payload.delta);
+          replace({ ...state, rotation: newRotation });
+        }
+      },
+    );
+
+    const unsubPitchBy = cameraBus.on(
+      CameraEventTypes.pitchBy,
+      ({ payload }) => {
+        if (payload.id !== mapId) {
```
colehart: same good defense here. Same request to add a test catching this:

```suggestion
        if (payload.id !== mapId || !Number.isFinite(payload.delta) {
```

packages/map-toolkit/src/camera/store.ts
Peechiz: good catch, that is a weird way to do it. fixing it now.

packages/map-toolkit/src/camera/camera.docs.mdx
colehart, line 463: You could add a section here noting the new pitchBy and rotateBy functions:

```suggestion
### Relative Adjustments

The hook also returns `pitchBy` and `rotateBy` actions for nudging the camera by a delta, without computing the absolute target yourself. These apply the same constraints as the events: `rotateBy` is ignored in 3D and normalizes to 0-360°, `pitchBy` only applies in 2.5D and clamps to 0-85°.

```tsx
const { pitchBy, rotateBy } = useMapCamera(MAP_ID);

// Tilt up 10° (clamped at 85°)
const handleTiltUp = () => pitchBy(10);

// Rotate 15° clockwise (wraps past 360°)
const handleRotateRight = () => rotateBy(15);
```
```

File: packages/map-toolkit/src/camera/store.ts (hunk 3 above)
Comment (colehart) [reply packages/map-toolkit/src/camera/store.ts:262]: actually - mapCursor does the same thing when emitting requestCursorChange, but you're right that this could work via a helper function. Something like: 

```tsx
function applyPitchDelta(state: CameraState, delta: number): CameraState {
  if (state.view !== '2.5D' || !Number.isFinite(delta)) return state;
  return { ...state, pitch: clamp(MIN_PITCH, MAX_PITCH, state.pitch + delta) };
}
// bus handler: replace(applyPitchDelta(get(), payload.delta))
// action:      replace(applyPitchDelta(get(), delta))
```

Peechiz:
I had the thought to explore adding an example of this new behavior to the `camera.stories.tsx`

something like

```typescript
    const dragHandler: BaseMapProps['onDrag'] = (_info, e) => {
      const isCtrlLeftDrag = e.srcEvent.ctrlKey && e.leftButton;

      if (e.rightButton || isCtrlLeftDrag) {
        if (cameraState.view === '2D') {
          setView({ id: CAMERA_STORY_ID, view: '2.5D' });
        }
        rotateBy(e.deltaX);
        pitchBy(-e.deltaY)
      }
    };

    return (
      <div className='relative h-dvh w-dvw'>
        <BaseMap
          className='absolute inset-0'
          id={CAMERA_STORY_ID}
          onDrag={dragHandler}
        />
        <CameraToolbar />
      </div>
    );
   ```
   
And even with the new events, it's still having the same sorts of issues with "snapping" to unintended pitch states (rotate is also broken, but differently - it rotates WAY too much)
   
When I perform a drag, the camera tilts down. I end the drag. When I go back in to make another adjustment, it seems to snap back to around 45 degrees. It _seems_ like it might be trying to apply a stale delta to the pitch somehow.

Rotation seems just _way_ too sensitive. Even trying to divide the delta by `10`, it seems to rotate way too much, and when I change directions it appears to lag a little.

colehart:
> I had the thought to explore adding an example of this new behavior to the `camera.stories.tsx`
> 
> something like
> 
> ```ts
>     const dragHandler: BaseMapProps['onDrag'] = (_info, e) => {
>       const isCtrlLeftDrag = e.srcEvent.ctrlKey && e.leftButton;
> 
>       if (e.rightButton || isCtrlLeftDrag) {
>         if (cameraState.view === '2D') {
>           setView({ id: CAMERA_STORY_ID, view: '2.5D' });
>         }
>         rotateBy(e.deltaX);
>         pitchBy(-e.deltaY)
>       }
>     };
> 
>     return (
>       <div className='relative h-dvh w-dvw'>
>         <BaseMap
>           className='absolute inset-0'
>           id={CAMERA_STORY_ID}
>           onDrag={dragHandler}
>         />
>         <CameraToolbar />
>       </div>
>     );
> ```
> 
> And even with the new events, it's still having the same sorts of issues with "snapping" to unintended pitch states (rotate is also broken, but differently - it rotates WAY too much)
> 
> When I perform a drag, the camera tilts down. I end the drag. When I go back in to make another adjustment, it seems to snap back to around 45 degrees. It _seems_ like it might be trying to apply a stale delta to the pitch somehow.
> 
> Rotation seems just _way_ too sensitive. Even trying to divide the delta by `10`, it seems to rotate way too much, and when I change directions it appears to lag a little.

Go ahead and push the story you wrote and we can all collab on fixing it with the same working context.

PR #1010, opened 2026-05-21
feat: add override icon to Notice

Peechiz:
Closes n/a
see: `CORE-177`

## ✅ Pull Request Checklist

## 📝 Test Instructions

Allows passing a custom icon into the `<Notice overrideIcon={<MyIcon />} />`

The automatic color-based switching is useful for basic notices, but shouldn't disallow customization for client apps.

- [x] Yes
- [ ] No

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [ ] Implementation

### DIFF
--- a/packages/design-toolkit/src/components/notice/index.tsx
+++ b/packages/design-toolkit/src/components/notice/index.tsx
@@ -73,6 +73,7 @@ export function Notice({
   primary,
   secondary,
   hideIcon,
+  iconOverride,
   showClose,
   shouldCloseOnAction,
   size = 'medium',
@@ -80,6 +81,9 @@ export function Notice({
   onSecondaryAction,
   onClose,
 }: NoticeProps) {
+  const showOverrideIcon = iconOverride && size === 'medium';
+  const showIcon = !hideIcon && size === 'medium' && !iconOverride;
+
   return (
     <Toast
       className={clsx('group/notice', styles.notice, classNames?.notice)}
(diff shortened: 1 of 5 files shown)


packages/design-toolkit/src/components/notice/index.tsx
```diff
@@ -73,13 +73,17 @@ export function Notice({
...
+  iconOverride,
   showClose,
   shouldCloseOnAction,
   size = 'medium',
   onPrimaryAction,
   onSecondaryAction,
   onClose,
 }: NoticeProps) {
+  const showOverrideIcon = iconOverride && size === 'medium';
+  const showIcon = !hideIcon && size === 'medium' && !iconOverride;
```
Peechiz, line 85: I'm unsure why we pass down `size` to `<NoticeIcon />` since it only ever renders at `medium` size. It's further confusing to see the `size={size === 'small' ? 'medium' : 'large'}` inside the NoticeIcon, since presumably this component never _gets_ a value other than `medium`

switzerb:
Yay! Thank you for picking up work. Here's the ticket: https://gitlab.accelint.dev/core-ux/standard-toolkit/-/issues/92

I think the solution you have in place may not be sufficient because mostly those Notice components are generated, and you won't be able to pass a function (React component) through the event bus. It has to be serializable.

packages/design-toolkit/src/components/notice/index.tsx
switzerb: There is a note about this in the ticket -- I believe it is just a bug (or AI hallucination) that missed the QA gate.

> There is an additional bug here where we only render the icon if the Notice size is medium, but in the Notice component we are checking to see if the size is small and then rendering a medium size icon – which makes no sense. The component will only render if the size is medium, so we can remove that conditional and just render the correct size of icon.

PR #1012, opened 2026-05-21
Fix: MGRS Parsing & Pasting Behavior

francistse23:
Closes 

#### Problem
During the recent NGC2 Lightning Surge 3 (LS3) event, we discovered the MGRS coordinate input/pasting could be improved to enhance UX.

#### Solution
This PR introduces a few changes:
- Uppercase normalization
- leading zero on blur
- paste normalization (casing, leading 0 padding)
- handle segment blur

We are also patching `geodesy` in the repo to use `Math.round` instead of `Math.floor` to prevent the loss of float point precision from the round trip conversion

Updated snapshot test that failed the visual regression test because of the decimal precision

## ✅ Pull Request Checklist

## 📝 Test Instructions

```
Coordinates to test with:

18S UU 82847 95355
6R YP 99663 76893
6R vp 35829 42741
06q xl 64413 35887
```

- Reference the above coordinates for testing
  - Other than the first coordinate, the other 3 should silently fail in the current build (i.e. pasting doesn't do anything, casing isn't normalized)
- Pull down this branch and run storybook
- Either type or paste in the values provided above
- You should see all of them work as intended
- Further inspect the story on screen, I have added a way for user to validate whether their input has been normalized

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [ ] Documentation
- [ ] Testing
- [x] Implementation

## 💬 Other information

Side by side comparison on the behavior. If you see nothing happen on the left side, it's because the pasting is silently failing

https://github.com/user-attachments/assets/a930231d-6647-459a-bd24-c0aacd7dea0b


packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-paste.ts
```diff
@@ -136,6 +136,31 @@ export function useCoordinatePaste({
...
+   * @param input - Raw MGRS coordinate string
+   * @returns Normalized MGRS coordinate string
+   *
+   * @example
+   * normalizeMgrsInput("6r yp 12345 67890") // → "06R YP 12345 67890"
+   * normalizeMgrsInput("6ryp1234567890")    // → "06RYP1234567890"
+   */
+  const normalizeMgrsInput = (input: string): string => {
+    // Convert to uppercase
+    let normalized = input.toUpperCase();
```
switzerb, line 154: super nit: I think in this case, the comment isn't really necessary. The `toUpperCase()` function sort of speaks for itself.

patches/geodesy@2.4.0.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -209,9 +209,9 @@ class Mgrs {
+ 
+         const { zone, band, e100k, n100k, easting, northing } = this;
+ 
+-        // truncate to required precision
+-        const eRounded = Math.floor(easting/Math.pow(10, 5-digits/2));
+-        const nRounded = Math.floor(northing/Math.pow(10, 5-digits/2));
++        // round to required precision (Math.round instead of Math.floor to handle floating-point precision errors)
++        const eRounded = Math.round(easting/Math.pow(10, 5-digits/2));
++        const nRounded = Math.round(northing/Math.pow(10, 5-digits/2));
```
switzerb: So I did a little research on this when we chatted about it (weeks? months? ago and I am not sure that rounding here is actually correct. Truncation (not rounding) actually appears to be an important element of managing precision in the MGRS grid system. My apologies for not circling back about this. https://en.wikipedia.org/wiki/Military_Grid_Reference_System#Truncation,_not_rounding

At the very least, I would like to validate this with an SME before we change the behavior of the library.

.changeset/small-corners-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@accelint/design-toolkit": minor
+---
+
+Fixed MGRS coordinates input behavior. Also, enhanced some of the UX with the function
```
switzerb, line 5: This would be a patch change for semantic versioning.

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-field-state.ts
```diff
@@ -272,6 +291,18 @@ export function useCoordinateFieldState({
     }
   };
 
+  const handleSegmentBlur = (index: number) => {
+    // MGRS zone field: add leading zero to single-digit values on blur
+    if (format === 'mgrs' && index === 0) {
+      const currentVal = segmentValues[0];
+      if (currentVal && currentVal.length === 1 && /^\d$/.test(currentVal)) {
+        const updatedValues = [...segmentValues];
+        updatedValues[0] = `0${currentVal}`;
```
switzerb: Might be cleaner and easier to read if you use the `padStart` function here instead of string concatenation.

packages/design-toolkit/src/hooks/coordinate-field/use-coordinate-paste.ts
```diff
@@ -136,6 +136,31 @@ export function useCoordinatePaste({
...
+
+    // Add leading zero to single-digit grid zones
+    // Pattern: single digit followed by letter at start (with optional space)
+    normalized = normalized.replace(/^(\d)([A-Z])/, '0$1$2');
```
switzerb: nit: same comment here -- up to you but I think since `padStart` is [designed to handle this exact use case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart), just makes the code a little more readable. not that anyone really reads code anymore.

switzerb (dismissed):

switzerb (approved):

PR #1023, opened 2026-06-01
fix: sidenav bugs

sotomaque:
#### Description
- Fixes four `Sidenav` visual bugs observed in JERIC2O, all isolated to the sidenav CSS module. 
- Each is a minimal, targeted change — no component restructuring. 

#### Changes

All in [`sidenav/styles.module.css`](packages/design-toolkit/src/components/sidenav/styles.module.css).

| Change | Bug / why |
| --- | --- |
| `.heading`: drop `uppercase` (`text-body-xs uppercase` → `text-body-xs`) | **Bug 1 — forced UPPERCASE headings.** The standalone `.heading` rule applied `text-transform: uppercase`, and the `text-body-*` utilities never reset it, so it reached *every* heading routed through `HeadingContext` — `SidenavMenu` titles, the popover title, `SidenavAvatar` headings, and `SidenavContent` section headings. Now renders verbatim. |
| `.transient`: `@apply hidden` → `@apply hidden!` | **Bug 2 — collapsed icons leak into the rail.** The `closed` variant wraps its condition in `:where()` (zero specificity), so the hide rule tied with `Icon`'s own `display: block` and lost on source order for icon children. Text hid; icons (the `SidenavLink` external arrow, `SidenavHeader` chevron) didn't. `!important` forces the collapsed-state hide to win for any child type. |
| `.menu .heading`: add `text-body-s` | **Bug 3 — undersized menu title.** The expanded trigger title rendered at `body-xs` (10px) — inherited section-label size — while its own collapsed popover title and sibling nav items are `body-s` (12px). It visibly shrank 12px→10px between states. Now `body-s` everywhere. |
| `.sidenav`: reorder `closed`/`group-push-left/layout`, pin push collapsed width to `var(--sidenav-width)` | **Bug 4 — push-mode rail ignored `--sidenav-width`.** The variable only affected overlay mode (consumed by the collapse transform); push layouts were content-driven, so the collapsed width was inconsistent and not themeable in push. Reordering also fixes a latent transform shift of the relative nav. |

#### Notes
- All headings now render verbatim. The only size variation is intentional: avatar headings 14px, nav text / menu title / popover title 12px, section labels 10px.
- Collapsed rail width is now governed by `--sidenav-width` (default 64px) in **both** overlay and push modes — a behavior change for consumers who set `--sidenav-width` to narrow only their overlay rail; it now applies to push too.
- `!important` on `.transient` is scoped to the collapsed state (the `closed` variant never matches when open), so it can't bleed into the expanded layout.

#### Test plan
- [ ] Storybook → **Components / Sidenav / Title Case And Collapsed Icons** (new story): renders expanded — confirm "The Watch" and avatar headings are title-case, and the menu title matches nav-item size. Collapse via the top-bar toggle → external-link arrow + chevron disappear cleanly, no leftover space.
- [ ] Existing **Default** story: collapse/expand → labels and arrows behave; section headings ("Title A", "External", "Menu") render verbatim.
- [ ] Push vs overlay: in a `DrawerLayout push="left"`, collapsed rail width now honors `--sidenav-width` and matches overlay.
- [ ] `pnpm --filter @accelint/design-toolkit test` → sidenav suite green (incl. new structural test asserting the link arrow carries `transient`).
- [ ] Verification gate: `pnpm build` · `pnpm lint` · `pnpm format` all pass.

### DIFF
--- a/packages/design-toolkit/src/components/sidenav/styles.module.css
+++ b/packages/design-toolkit/src/components/sidenav/styles.module.css
@@ -30,17 +30,6 @@
     @variant group-push-left/layout {
       @apply relative;
       transform: translateX(0);
-
-      /*
-       * Push keeps the nav in the layout flow, so the overlay translate can't
-       * reveal a fixed rail. Pin the collapsed width to --sidenav-width so the
-       * collapsed rail matches overlay mode and stays themeable from one
-       * variable. (Declared after `closed` so push's translateX(0) wins and the
-       * relative nav isn't shifted.)
-       */
-      @variant closed {
-        width: var(--sidenav-width);
-      }
     }
   }
 
--- a/packages/design-toolkit/src/components/sidenav/styles.module.css
+++ b/packages/design-toolkit/src/components/sidenav/styles.module.css
@@ -74,7 +74,7 @@
     }
 
     .heading {
-      @apply fg-primary-bold text-body-m text-left;
+      @apply fg-primary-bold text-body-m text-left normal-case;
       grid-area: heading;
     }
 
@@ -84,8 +84,14 @@
     }
   }
 
+  /*
+   * Section labels (e.g. "Title A", "External" — the DEFAULT_SLOT headings
+   * inside SidenavContent) render uppercase per the design system. Headings
+   * routed to the menu/panel/avatar contexts opt out via `text-transform: none`
+   * in their own nested `.heading` rules below.
+   */
   .heading {
-    @apply text-body-xs;
+    @apply text-body-xs uppercase;
   }
 
   .item {
@@ -172,7 +178,7 @@
     @apply text-body-s relative;
 
     .heading {
-      @apply text-body-s flex-1 text-left;
+      @apply text-body-s flex-1 text-left normal-case;
     }
 
     .item {
@@ -256,7 +262,7 @@
     @apply fg-primary-bold rounded-medium bg-surface-raised outline-static outline;
 
     .heading {
-      @apply rounded-t-medium bg-surface-overlay px-m py-s text-body-s;
+      @apply rounded-t-medium bg-surface-overlay px-m py-s text-body-s normal-case;
     }
 
     .content {
@@ -305,15 +311,8 @@
   }
 
   .transient {
-    /*
-     * `hidden!` is intentional: the `closed` variant wraps its condition in
-     * `:where()` (zero specificity), so this rule ties with `.icon`'s own
-     * `display: block` and loses on source order for icon children (e.g. the
-     * SidenavLink external-link arrow and the SidenavHeader collapse chevron).
-     * `!important` guarantees the hide wins for any child type when collapsed.
-     */
     @variant group-closed/sidenav {
-      @apply hidden!;
+      @apply hidden;
     }
   }
 }
--- a/packages/design-toolkit/src/components/sidenav/styles.module.css
+++ b/packages/design-toolkit/src/components/sidenav/styles.module.css
@@ -22,14 +22,25 @@
         var(--animation-easing-standard);
     }
 
-    @variant group-push-left/layout {
-      @apply relative;
-      transform: translateX(0);
-    }
-
     @variant closed {
       @apply items-center;
       transform: translateX(calc(-100% + var(--sidenav-width)));
+    }
+
+    @variant group-push-left/layout {
+      @apply relative;
+      transform: translateX(0);
+
+      /*
+       * Push keeps the nav in the layout flow, so the overlay translate can't
+       * reveal a fixed rail. Pin the collapsed width to --sidenav-width so the
+       * collapsed rail matches overlay mode and stays themeable from one
+       * variable. (Declared after `closed` so push's translateX(0) wins and the
+       * relative nav isn't shifted.)
+       */
+      @variant closed {
+        width: var(--sidenav-width);
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/design-toolkit/src/components/sidenav/styles.module.css
```diff
@@ -74,7 +85,7 @@
   }
 
   .heading {
-    @apply text-body-xs uppercase;
+    @apply text-body-xs;
   }
```
shadyendless, line 84: Not letting me suggest the code change, but this can be:
```css
  .heading {
    @apply text-body-xs uppercase;

    * {
      text-transform: unset;
    }
  }
```

It keeps the uppercase and resets it for any children.

packages/design-toolkit/src/components/sidenav/styles.module.css
```diff
@@ -161,7 +172,7 @@
     @apply text-body-s relative;
 
     .heading {
-      @apply flex-1 text-left;
+      @apply text-body-s flex-1 text-left;
```
shadyendless: I don't fully understand this change/can't seem to replicate what the changelog said it was fixing. Can you provide an example so I can better understand what it was implying?

packages/design-toolkit/src/components/sidenav/styles.module.css
```diff
@@ -294,8 +305,15 @@
...
+    /*
+     * `hidden!` is intentional: the `closed` variant wraps its condition in
+     * `:where()` (zero specificity), so this rule ties with `.icon`'s own
+     * `display: block` and loses on source order for icon children (e.g. the
+     * SidenavLink external-link arrow and the SidenavHeader collapse chevron).
+     * `!important` guarantees the hide wins for any child type when collapsed.
+     */
     @variant group-closed/sidenav {
-      @apply hidden;
+      @apply hidden!;
```
shadyendless: I am not sure this is necessary. Removing it locally shows no difference in behavior when comparing my local preview to the deployed preview for this PR. Do you see different behavior when running locally?

packages/design-toolkit/src/components/sidenav/styles.module.css
sotomaque: `text-transform: unset` is a no-op here — text-transform is an inherited property, so unset resolves to inherit, which re-applies the parent's uppercase. (It would need  `text-transform: none`.)

Even with none, a child selector can't fix the main manifestation: these headings are plain text content (`<Heading>The Watch</Heading>`  -> text node directly inside the h3), so there's no child element to reset, the heading's own text stays uppercase.

packages/design-toolkit/src/components/sidenav/styles.module.css
sotomaque: Concrete repro on main: Storybook → Components/Sidenav → Default → expand the rail → inspect the menu trigger title ("Nav Item" under the "Menu" section heading). Computed font-size is 10px — it falls through to the standalone .heading { text-body-xs } rule because .menu .heading declared no size. Compare: the sibling nav-item labels ("Nav item") are 12px (.text is body-s), and the same menu's collapsed popover title is 12px (.panel .heading sets body-s). So the title visibly changes 12px → 10px between collapsed-popover and expanded states, and  sits undersized next to its sibling items. The text-body-s addition pins it to 12px in both states.

packages/design-toolkit/src/components/sidenav/styles.module.css
sotomaque: You're not seeing a difference because the tie currently happens to resolve in our favor in your build — but it's a genuine tie, which is the problem. 

Both selectors are specificity 0,1,0: the closed variant compiles to `._transient:is(:where(.group/sidenav):where(:not(...)) *)` (everything in `:where()` contributes zero), and Icon's own display: block is `._icon { …  }`. 

A tie is decided by rule order across compiled CSS-module files in the final bundle, which depends on the consumer's import graph/chunking — not something we control or that's  stable across apps (this is the same zero-specificity ordering class as the push-mode bug in this PR, and the leak was observed in JERIC2O's bundle). 

The `!important` makes the collapsed hide deterministic regardless of bundle order.

If we'd rather avoid !important: doubling the class (&.transient) bumps the rule to 0,2,0, which deterministically beats Icon's 0,1,0 without relying on order. Happy to switch to that.

packages/design-toolkit/src/components/sidenav/styles.module.css
shadyendless: So the intention here is to not have the navigation headings (unsure what to call them) not be uppercase unless explicitly opted into?

I think the design system had it so that those remained uppercase, is there a different use-case being presented in your project?

packages/design-toolkit/src/components/sidenav/styles.module.css
shadyendless: I am still not convinced this is necessary with the changes to fix the width.

shadyendless (dismissed):

packages/design-toolkit/src/components/sidenav/styles.module.css
```diff
@@ -22,14 +22,25 @@
...
+      /*
+       * Push keeps the nav in the layout flow, so the overlay translate can't
+       * reveal a fixed rail. Pin the collapsed width to --sidenav-width so the
+       * collapsed rail matches overlay mode and stays themeable from one
+       * variable. (Declared after `closed` so push's translateX(0) wins and the
+       * relative nav isn't shifted.)
+       */
+      @variant closed {
+        width: var(--sidenav-width);
+      }
```
shadyendless: I think this is breaking it.

shadyendless (approved):

PR #1026, opened 2026-06-09
Feat/ GPU-accelerated map camera transitions

ipetez:
⏺ Closes <!-- No corresponding GitHub issue -->
                                                                                                                                  
  ## ✅ Pull Request Checklist                                                                                                      
                                                                                                                                    
  - [x] The commit message follows [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary)                    
  [extended](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) guidelines.                                      
                                                                                                                                    
  ## 📝 Test Instructions                                                                                                           
                                                                                     
  1. Run Storybook: `pnpm dev:storybook`                                                                                            
  2. Navigate to **Camera > Smooth Transitions** story
  3. Test location transitions:                                                                                                     
     - Click any US city button (New York, Los Angeles, Chicago, Miami, Seattle)                                                    
     - Observe smooth animated camera movement with 2s ease-out transition                                                          
     - Click multiple cities in sequence to verify smooth transitions work consistently                                             
  4. Test different easing functions:                                                                                               
     - Use the "Easing Functions" buttons to fly to Chicago with different animations                                               
     - Compare Linear, Ease In, Ease Out, and Ease In-Out (all 3s duration)                                                         
  5. Verify current position display updates correctly at the bottom of the toolbar                                                 
                                                                                                                                    
  Expected behavior: Camera should smoothly animate between locations using MapLibre's `easeTo()` method.  

<img width="1430" height="620" alt="Jun-09-2026 14-11-40" src="https://github.com/user-attachments/assets/98dec2ee-6c44-4f29-8de8-27532986d0b4" />
                                                                                                  
                                                                                                                                    
  - [ ] Yes                                                                                                                         
  - [x] No                                                                           
                                                 
  ## 🤖 AI Usage                                                                                                                    
   
  - [x] Documentation                                                                                                               
  - [x] Testing                                                                      
  - [x] Implementation                                                                                                              
                               
  AI tool used: Claude Code                                                                                                         
                                                                                     
  ## 💬 Other information                                                                                                           
                                                                                     
  This feature adds smooth camera transitions to the map-toolkit camera system by:                                                  
  - Adding `transitionDuration` and `transitionEasing` properties to `CameraSetCenterEvent` payload
  - Implementing MapLibre's `easeTo()` method in BaseMap component for animated camera movements                                    
  - Supporting four easing functions: linear, ease-in, ease-out, ease-in-out                                                        
  - Providing an interactive Storybook demo to showcase the functionality                                                           
                                                                                                                                    
  The implementation ensures transitions work correctly by managing transition state and temporarily excluding camera position from 
  mapOptions during animations to prevent conflicts between React state updates and MapLibre's animation system.


packages/map-toolkit/src/camera/types.ts
```diff
@@ -48,6 +48,10 @@ export type CameraSetCenterPayload = {
   heading?: number;
   /** Optional pitch angle */
   pitch?: number;
+  /** Optional transition duration in milliseconds */
+  transitionDuration?: number;
+  /** Optional transition easing function */
+  transitionEasing?: 'linear' | 'ease-in' | 'ease-out' | 'ease-in-out';
```
switzerb: since these four strings are repeated in multiple places, let's make them a first class TransitionEasing (or just Easing) type
```suggestion
  transitionEasing?: TransitionEasing;
```

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -415,6 +440,64 @@ export function BaseMap({
...
+        setCameraState({
+          latitude: cameraState.latitude,
+          longitude: cameraState.longitude,
+          zoom: cameraState.zoom,
+          pitch: cameraState.pitch,
+          rotation: cameraState.rotation,
+          transitionDuration: undefined,
+          transitionEasing: undefined,
+        });
+      }, cameraState.transitionDuration + 50);
```
switzerb: I have some concern about the `setTimeout` here as the way to end the animation -- because if something is amiss then we will leave the camera store out of sync (line 400). I really like the overall approach and I think using maplibre's easeTo is the right approach.  [In the docs](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#easeto) it says that the easeTo function triggers an `moveend` event -- maybe instead of a setTimeout to clear the animation duration we can use `map.once('moveend')? I wish there was a callback. 

I also think we are assuming here that there will only ever be one transition in flight and that it won't get interrupted -- but I can totally see a user grabbing the screen mid-transition and trying to drag to a new place. If that happens and the timeout fires, the user will get snapped back to the original destination, I think. If we return a `map.stop()` or some other cleanup from the `useEffect` here then we can kill whatever was in flight previously. Pun absolutely intended.

packages/map-toolkit/src/camera/types.ts
ipetez: Makes sense! Updated ✅

packages/map-toolkit/src/deckgl/base-map/index.tsx
ipetez: Awesome! Yeah subscribing to the `moveend` event is much better here since it avoids potential race conditions  but also, to your point, handles interruptions appropriately so now if the user drags the map mid-transition, moveend fires immediately when they stop. Let me know if the refactor makes sense to you.

switzerb (approved):

PR #1027, opened 2026-06-10
fix: notice list bugs

sotomaque:
## Summary

  Fixes four bugs in `NoticeList` (`packages/design-toolkit/src/components/notice/list.tsx`):
  
  1. **"Clear All" crashes and does nothing** — `onPress={queue.clear}` passed the method unbound; react-stately's `ToastQueue.clear()` references `this.queue`, so clicking threw `TypeError: Cannot set properties of undefined (setting
  'queue')` and the list was never cleared.
  2. **List-level defaults override per-notice values** — `defaultTimeout ?? payload.timeout` and `defaultColor || payload.color` gave the *list* precedence, contradicting the documented behavior ("Default color for notices without explicit
  color"). A notice emitted with `color: 'critical'` rendered with the list's `defaultColor` instead.
  3. **Untargeted lists consume targeted notices** — a `NoticeList` without an `id` accepted *every* queue event, including ones explicitly `target`ed at another list, so targeted notices rendered twice when multiple lists were mounted.
  4. **Queue subscription leak** — the `useEffect` had no dependency array and no cleanup, adding a new `queue.subscribe` listener on every render and never removing any.
  
  ## Repro on `main`

  1. `pnpm --filter @accelint/design-toolkit preview` → open Storybook
  2. **Crash:** `Components/NoticeList → Default` → set `hideClearAll: false` in controls → click **Create Notice** twice → click **Clear All** → console shows `TypeError: Cannot set properties of undefined (setting 'queue')`, notices remain
  on screen.
  3. **Precedence:** render a `NoticeList` with `defaultColor='info'` and emit `Notice:queue` with `color: 'critical'` → the notice renders info-colored.
  4. **Targeting:** render two lists — one with an `id`, one without — and emit a notice with `target` set to the first list's id → it appears in **both** lists.
  5. **Leak:** code inspection — `list.tsx` `useEffect` with no deps/cleanup calls `queue.subscribe` every render.

  ## Fix

  - `onPress={() => queue.clear()}`
  - `payload.timeout ?? defaultTimeout` and `payload.color || defaultColor` (per-notice wins; defaults fill gaps)
  - Queue guard: `id ? payload.target === id : !payload.target` — an id-less list only accepts untargeted notices
  - `useEffect(() => queue.subscribe(...), [queue])` — `subscribe()` returns its unsubscribe function, now used as the effect cleanup

  ## Test plan

  - [ ] New regression tests in `list.test.tsx`:
    - Clear All removes all notices without throwing
    - per-notice `color` wins over `defaultColor`
    - `defaultColor` still applies when the notice has no color
    - id-less list ignores notices targeted at another list
    - id'd list receives only its own targeted notices
  - [ ] Existing timeout tests still pass (precedence flip preserves default-only and payload-only behavior)
  - [ ] `pnpm build && pnpm test && pnpm lint && pnpm format`

### DIFF
--- a/packages/design-toolkit/src/components/notice/list.tsx
+++ b/packages/design-toolkit/src/components/notice/list.tsx
@@ -111,13 +111,13 @@ export function NoticeList({
   const emitClose = useEmit(NoticeEventTypes.close);
 
   useOn(NoticeEventTypes.queue, (data) => {
-    if ((id && data.payload.target === id) || !id) {
-      const timeout = defaultTimeout ?? data.payload.timeout;
+    if (id ? data.payload.target === id : !data.payload.target) {
+      const timeout = data.payload.timeout ?? defaultTimeout;
       queue.add(
         {
           ...data.payload,
           id: data.payload.id || uuid(),
-          color: defaultColor || data.payload.color,
+          color: data.payload.color || defaultColor,
         },
         {
           timeout,
@@ -146,11 +146,13 @@ export function NoticeList({
     }
   });
 
-  useEffect(() => {
-    queue.subscribe(() => {
-      setHasNotices(queue.visibleToasts.length > 0);
-    });
-  });
+  useEffect(
+    () =>
+      queue.subscribe(() => {
+        setHasNotices(queue.visibleToasts.length > 0);
+      }),
+    [queue],
+  );
 
   const children = (
     <>
@@ -161,7 +163,7 @@ export function NoticeList({
             (className) => className ?? '',
           )}
           variant='outline'
-          onPress={queue.clear}
+          onPress={() => queue.clear()}
         >
           Clear All
         </Button>
--- a/.changeset/fix-notice-list-queue-handling.md
+++ b/.changeset/fix-notice-list-queue-handling.md
@@ -1,0 +1,10 @@
+---
+'@accelint/design-toolkit': patch
+---
+
+Fix four `NoticeList` bugs:
+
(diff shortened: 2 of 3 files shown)


packages/design-toolkit/src/components/notice/list.tsx
```diff
@@ -111,13 +111,13 @@ export function NoticeList({
...
-    if ((id && data.payload.target === id) || !id) {
-      const timeout = defaultTimeout ?? data.payload.timeout;
+    if (id ? data.payload.target === id : !data.payload.target) {
+      const timeout = data.payload.timeout ?? defaultTimeout;
       queue.add(
         {
           ...data.payload,
           id: data.payload.id || uuid(),
-          color: defaultColor || data.payload.color,
+          color: data.payload.color || defaultColor,
```
shadyendless, line 120: Should this also be `??` rather than `||` or is this to mitigate against empty strings?

sotomaque: Intentional! `color` is a string enum arriving over the event bus, where emitters aren't statically type-checked end-to-end. With `??`, an empty-string (or other falsy junk) color would flow through to `<Notice color=''>` and render `data-color=""` with no severity styling; makes anything falsy fall back to `defaultColor`.

For the statically-valid cases (undefined vs a real color) `||` and `??` behave identically here.

shadyendless (approved):

PR #1028, opened 2026-06-10
fix: coordinate field

sotomaque:
## Summary

  Two `CoordinateField` correctness fixes:

  1. **Controlled mode wipes the whole field mid-edit** (`@accelint/design-toolkit`) — the hook's value-sync effect re-ran on every `value` prop change. Emptying one segment (or typing an
  invalid coordinate) emits `onChange(null)`; a controlled parent echoes that `null` straight back, and the effect rebuilt **all** segments from it — clearing the longitude when you
  backspaced the latitude, and erasing the "Invalid coordinate value" error one render after it appeared.
  2. **DDM/DMS formatting produces impossible "60" minutes/seconds** (`@accelint/geo`) — `toFixed` rounding in `formatDegreesDecimalMinutes`/`formatDegreesMinutesSeconds` could roll
  minutes/seconds to exactly 60 with no carry, so `40.9999995` formatted as `40° 60.0000'` instead of `41° 0.0000'`. The field then flagged its own output as invalid on blur.

  ## Repro on `main`

  1. `pnpm --filter @accelint/design-toolkit preview` → open Storybook
  2. **Controlled wipe:** `Components/CoordinateField → With Initial Value` → click the latitude segment → press Backspace → **both** latitude and longitude clear (`["40.7128","-74.006"]`
  → `["",""]`). Or type `95` as latitude → error flashes, then the whole field empties.
  3. **60-minutes:** render `<CoordinateField format='ddm' value={{ lat: 40.9999995, lon: -74.006 }} />` → segments read `40° 60' N` — an impossible coordinate. Same root cause in
  `@accelint/geo`: `formatDegreesDecimalMinutes([40.9999995, -74.006])` → `"40° 60.0000', 74° 0.3600'"`.

  ## Fix

  - **Echo guard** (`use-coordinate-field-state.ts`): the hook records the last value it emitted via `onChange` (`lastEmittedValueRef`). The value-sync effect now skips re-syncing when the
   incoming controlled `value` is that same echo (compared by `lat`/`lon`, with `null === null`), so in-progress edits and validation errors survive. Genuinely new external values still
  re-sync as before.
  - **Rounding carry** (`geo` formatters): after rounding, `minutes >= 60` carries into degrees (DDM) and `seconds >= 60` carries into minutes, then minutes into degrees (DMS). Output
  precision/format unchanged.

  ## Test plan

  - [ ] New `formatter.test.ts` for DDM + DMS in `@accelint/geo`: typical values unchanged, carry at `40.9999995`, double carry (`59' 60.00″` → `+1°`), pole/antimeridian boundary
  (`89.99999999` → `90° 0'`)
  - [ ] New controlled-mode regression tests in `coordinate-field.test.tsx`: clearing one segment leaves the other intact; the validation error stays visible (and segments keep the typed
  text) after an invalid entry
  - [ ] Existing coordinate-field + geo parser suites still pass
  - [ ] `pnpm build && pnpm test && pnpm lint && pnpm format`

### DIFF
--- a/packages/geo/src/coordinates/latlon/degrees-minutes-seconds/formatter.ts
+++ b/packages/geo/src/coordinates/latlon/degrees-minutes-seconds/formatter.ts
@@ -31,12 +31,23 @@ import { createFormatter } from '../internal/format';
  * ```
  */
 const toDegreesMinutesSeconds = (num: number): string => {
-  const degrees = Math.floor(Math.abs(num));
+  let degrees = Math.floor(Math.abs(num));
   const minutesFull = (Math.abs(num) - degrees) * 60;
-  const minutes = Math.floor(minutesFull);
-  const seconds = ((minutesFull - minutes) * 60).toFixed(2);
+  let minutes = Math.floor(minutesFull);
+  let seconds = Number(((minutesFull - minutes) * 60).toFixed(2));
 
-  return `${degrees}° ${minutes}' ${seconds}″`;
+  // Rounding can produce 60 seconds (e.g. 40.9999999 -> 40° 59' 60.00″);
+  // carry into minutes (and degrees) so the output stays a valid coordinate.
+  if (seconds >= 60) {
+    minutes += 1;
+    seconds = 0;
+  }
+  if (minutes >= 60) {
+    degrees += 1;
+    minutes = 0;
+  }
+
+  return `${degrees}° ${minutes}' ${seconds.toFixed(2)}″`;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


packages/geo/src/coordinates/latlon/degrees-minutes-seconds/formatter.ts
```diff
@@ -31,12 +31,23 @@ import { createFormatter } from '../internal/format';
...
+  // Rounding can produce 60 seconds (e.g. 40.9999999 -> 40° 59' 60.00″);
+  // carry into minutes (and degrees) so the output stays a valid coordinate.
+  if (seconds >= 60) {
+    minutes += 1;
+    seconds = 0;
+  }
+  if (minutes >= 60) {
+    degrees += 1;
+    minutes = 0;
+  }
```
shadyendless:
```suggestion
minutes += Math.floor(seconds / 60);
seconds %= 60;

degrees += Math.floor(minutes / 60);
minutes %= 60;
```

shadyendless (approved):

PR #1032, opened 2026-06-15
feat: mouse camera controls in 2.5D

colehart:
Closes issue raised in neo demo

## ✅ Pull Request Checklist

## 📝 Test Instructions

- head to http://localhost:6007/?path=/story/camera--basic-usage after checking out this branch
- left click to pan and notice in 2 and 3D, right click does nothing
- switch to 2.5D and right click to pitch and rotate the camera as well as left click to pan

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [x] Testing
- [x] Implementation


packages/map-toolkit/src/camera/camera.stories.tsx
```diff
@@ -93,7 +93,7 @@ function CameraToolbar() {
           value={cameraState.pitch}
           isDisabled={cameraState.view !== '2.5D'}
           minValue={0}
-          maxValue={65}
+          maxValue={85}
```
switzerb: Maybe not for this PR, but it would be useful to export some of the constants from map-toolkit for this kind of scenario in an application, so users can set a slider to the maxpitch as a constant that they know will work with the camera, rather than having to know a magic number.

packages/map-toolkit/src/deckgl/base-map/index.tsx
```diff
@@ -512,7 +673,13 @@ export function BaseMap({
...
           key={mapGeneration}
           onMove={(evt) => {
-            setCameraState(evt.viewState);
+            // MapLibre reports heading as `bearing`, but the camera store keys on
+            // `rotation` (and `buildCameraState` ignores `bearing`). Without this
+            // remap, a right-drag rotation is dropped on the round-trip and the
+            // viewState memo pushes the stale `rotation` back to the map, snapping
+            // the bearing back. Pitch is unaffected since it's named the same.
+            const { bearing, ...rest } = evt.viewState;
+            setCameraState({ ...rest, rotation: bearing });
```
switzerb, line 685: This might not be an issue, but here we are mapping bearing to rotation on all move events, not just dragging / tilting. If there are other move events that set bearing, then they will update rotation. That might be what we want? (I am not familiar enough with all the move events and their properties for MapLibre...but just calling it out here.

switzerb (dismissed):
couple of small comments but nothing to block merge I think.

packages/map-toolkit/src/deckgl/base-map/index.tsx
colehart: I'll beef up the comment to explain that more - but yes - we need to keep everything synced

packages/map-toolkit/src/camera/camera.stories.tsx
colehart: I'll export the MAX_PITCH

packages/map-toolkit/src/deckgl/base-map/index.tsx
switzerb: thank you -- much clearer

switzerb (approved):

PR #1048, opened 2026-07-03
1047 pass metadata to notice onprimaryaction callback

### LINKED ISSUE #1047: Pass metadata to notice onPrimaryAction callback
### Provide a general summary of the feature

 Add direct callback props to NoticeList for action handling without requiring bus events. 

### 🤔 Expected Behavior?

`NoticeList` should accept optional callback props that are invoked when action buttons are pressed, with metadata passed as a parameter:

```ts
  function MyComponent() {
    const centerOn = useCenterOn();

    const handlePrimaryAction = (metadata) => {
      if (metadata?.geometry) {
        centerOn({ geometry: metadata.geometry, zoomTo: true });
      }
    };

    return (
      <NoticeList 
        id="map-notices"
        placement="top right"
        onPrimaryAction={handlePrimaryAction}
        onSecondaryAction={handleSecondaryAction}
        onClose={handleClose}
      />
    );
  }
```

### 😢 Current Behavior
  `NoticeList` only supports action handling through bus events. You must use useOn to listen for action events:

```ts
  function MyComponent() {
    const centerOn = useCenterOn();

    useOn(NoticeEventTypes.actionPrimary, (event) => {
      const { metadata } = event.payload;
      if (metadata?.geometry) {
        centerOn({ geometry: metadata.geometry, zoomTo: true });
      }
    });

    return <NoticeList id="map-notices" placement="top right" />;
  }
```

This works but adds indirection for simple, component-scoped action handling.

### 🤩 Possible Solution

Add optional callback props to NoticeList:

```ts
type NoticeListProps = {
    // ... existing props
    onPrimaryAction?: (metadata?: NoticeContent['metadata']) => void;
    onSecondaryAction?: (metadata?: NoticeContent['metadata']) => void;
    onClose?: (id: UniqueId, metadata?: NoticeContent['metadata']) => void;
};
```
Implementation would:
  1. Call these callbacks (if provided) when actions occur
  2. Still emit bus events (maintaining backward compatibility)
  3. Pass metadata from the notice to the callback

This gives developers the choice between direct callbacks (simpler for local handling) or bus events (better for global/cross-component scenarios).

### 📜 Context

I am emitting a queue event when a certain condition is met and want to include the data that meets said condition. 

I render notices with a message and a "Center On" button. The action (centering the map) is only relevant to the map component that displayed the notice. Currently I have to intercept the queue event and store a map of entities that are passed along in the metadata., creating unnecessary indirection.

 For component-scoped actions, direct callbacks are more ergonomic and easier to understand similar to how Button works with onPress rather than requiring event listeners.

### 💻 Examples

  Use Case 1: Map layer errors with Center On action
```ts
function MapView() {
    const centerOn = useCenterOn();
    const emit = useEmit(NoticeEventTypes.queue);

    const showLayerError = (layer) => {
      emit({
        message: `Layer "${layer.name}" failed to load`,
        primary: { children: "Center On" },
        metadata: { geometry: layer.geometry }
      });
    };

    return (
      <NoticeList 
        id="map-notices"
        onPrimaryAction={(metadata) => {
          centerOn({ geometry: metadata.geometry, zoomTo: true });
        }}
      />
    );
  }
```

  Use Case 2: File upload with retry
```ts  
function FileUploader() {
    const retryUpload = useRetryUpload();

    return (
      <NoticeList 
        id="upload-notices"
        onPrimaryAction={(metadata) => {
          if (metadata?.fileId) {
            retryUpload(metadata.fileId);
          }
        }}
      />
    );
  }
```

hendrix-io:
Closes #1047 

## ✅ Pull Request Checklist

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [x] Documentation
- [ ] Testing
- [x] Implementation

### DIFF
--- a/packages/design-toolkit/src/components/notice/utils.ts
+++ b/packages/design-toolkit/src/components/notice/utils.ts
@@ -14,24 +14,49 @@
 
 import 'client-only';
 import { isEqual } from 'radashi';
+import type { NoticeContent, NoticeDequeueEvent } from './types';
 
 /**
- * Checks if a payload matches metadata on a notice.
- * Used for filtering notices during dequeue operations.
+ * Checks if a dequeue payload matches a notice's content.
+ * Supports filtering by top-level fields (id, color, target) and nested metadata.
  *
- * @param payload - Object with key-value pairs to match against.
- * @param metadata - Metadata object from a notice to check.
- * @returns True if all payload entries match corresponding metadata entries.
+ * @param dequeuePayload - Dequeue event payload with optional id, color, target, metadata.
+ * @param noticeContent - Notice content to match against.
+ * @returns True if all provided criteria match.
  */
-export function matchesMetadata(
-  payload: Record<string, unknown>,
-  metadata?: Record<string, unknown>,
+export function matchesDequeueFilter(
+  dequeuePayload: NoticeDequeueEvent['payload'],
+  noticeContent: NoticeContent,
 ) {
-  if (!metadata) {
+  // Match top-level id
+  if (dequeuePayload.id && dequeuePayload.id !== noticeContent.id) {
     return false;
   }
 
-  return Object.entries(payload).every(
-    ([key, value]) => key in metadata && isEqual(metadata[key], value),
+  // Match top-level target
+  if (dequeuePayload.target && dequeuePayload.target !== noticeContent.target) {
+    return false;
+  }
+
+  // Match top-level color
+  if (dequeuePayload.color && dequeuePayload.color !== noticeContent.color) {
+    return false;
+  }
+
+  const dequeueMetadata = dequeuePayload.metadata;
+  const noticeMetadata = noticeContent.metadata;
+
+  // Match metadata fields (partial deep match)
+  if (!dequeueMetadata) {
+    return true;
+  }
+
+  if (!noticeMetadata) {
+    return false;
+  }
+
+  return Object.entries(dequeueMetadata).every(
+    ([key, value]) =>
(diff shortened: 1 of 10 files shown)


packages/design-toolkit/src/components/notice/utils.ts
```diff
@@ -14,24 +14,49 @@
...
- * @param metadata - Metadata object from a notice to check.
- * @returns True if all payload entries match corresponding metadata entries.
+ * @param dequeuePayload - Dequeue event payload with optional id, color, target, metadata.
+ * @param noticeContent - Notice content to match against.
+ * @returns True if all provided criteria match.
  */
-export function matchesMetadata(
-  payload: Record<string, unknown>,
-  metadata?: Record<string, unknown>,
+export function matchesDequeueFilter(
```
switzerb, line 48: while I am pretty sure it is unlikely that anyone is directly using this util function, it *is* a part of the API of the design-toolkit so if anyone is importing `import {matchesMetadata} from '@accelint/design-toolkit'` it makes this update a breaking change. Also, the signature of the function is changing so even if we keep the name the same, it's still "break-y".

But the PR makes sense and seems useful, so maybe we can just extract the common logic for the two functions and then add in the additional conditionals to the new function and just leave the other `matchesMetadata` function with a deprecated tag, to avoid shipping breaking changes.

hendrix-io: yeah good call!

hendrix-io: addressed!

switzerb (approved):
LGTM

PR #1059, opened 2026-08-13
Fix/temporal cancel next second

jemccarthy13:
Closes  [No issue to link]

callNextSecond sets a timeout and (on main) does not return a cleanup.

setClockInterval uses callNextSecond to align the callback with the next clock second. 

If setClockInterval is invoked at 001 ms, there are 999 ms where the callback is in queue for the repeat call.

If, at 500 ms, the application calls setClockIntervals cleanup, callNextSecond is not effectively cleaned up, and the repeat() call will revive the original callNextSecond timeout.

This MR: updates callNextSecond to return a cleanup function, such that setClockInterval can clean up the original callNextSecond timeout.

## ✅ Pull Request Checklist

- [?] Included link to corresponding [GitHub Issue](https://github.com/gohypergiant/standard-toolkit/issues).
- [?] The commit message follows [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) [extended](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) guidelines.

## 📝 Test Instructions

- Run the packages/temporal unit tests

- [ ] Yes
- [x] No

## 🤖 AI Usage

- [ ] Documentation
- [x] Testing
- [ ] Implementation

ChatGPT was used to triage the issue and create the test. Solution implemented by human.


packages/temporal/src/timers/index.ts
```diff
@@ -43,9 +43,12 @@ export function setClockInterval(cb: () => void, ms: number) {
...
   }
 
-  callNextSecond(repeat);
+  const cancelNextSecond = callNextSecond(repeat);
 
-  return () => clearTimeout(timeout);
+  return () => {
+    cancelNextSecond();
+    if (timeout !== undefined) clearTimeout(timeout);
+  };
```
devindavies, line 51: I'm not sure this covers an important edge case, the callback will fire after cleanup if cleanup is called within the callback. You can prove this with a test like this:

```
 it('honors cleanup when called from inside the callback', () => {
      vi.spyOn(Date, 'now').mockReturnValue(13_000); // Start at exact second

      let cleanup: (() => void) | undefined;

      const cb = vi.fn(() => {
        // Call cleanup from inside the callback
        if (cleanup) {
          cleanup();
        }
      });

      cleanup = setClockInterval(cb, 1000);

      // Advance to trigger first call at next second boundary
      vi.advanceTimersByTime(1000);
      expect(cb).toHaveBeenCalledTimes(1);

      // Callback should NOT fire again since cleanup was called
      vi.advanceTimersByTime(1000);
      expect(cb).toHaveBeenCalledTimes(1);

      // Even after multiple intervals
      vi.advanceTimersByTime(5000);
      expect(cb).toHaveBeenCalledTimes(1);
    });
```

To fix this you may need a `cancelled` boolean flag for early returning from `repeat()`

packages/temporal/src/timers/index.ts
devindavies, line 72: we should address this issue in setClockTimeout as well

devindavies:
please add a changeset

devindavies (approved):

PR #1065, opened 2026-08-27
chore: allow undefined values in lookup

jemccarthy13:
- Allow null | undefined in lookup function calls.

You can see the TS error if you add the explicit undefined params test cases to 'main'.

The 
`const fn = def ?? identity` 
move is to lift that assignment to the time of creation, rather than time of call

```typescript
const TABLE = {
 'FIRST': 'HELLO',
 'SECOND': 'WORLD'
}

const tableLookup = lookup(TABLE, x => x ?? 'MISSING')

/* ------------------- */

// Let's assume 'value' corresponds to "FIRST" or "SECOND"
// but it's also nullable
type MyData = {
  value?: string,
  id: string,
}

// we get this data from an external source; if the source says nullable
// data.value could be undefined|null
const data: MyData = fetch(...)

//    a representative/mocked return might be:
// const mockMyData: MyData = {
//   id: '1234',
// }
//


// before this PR - ts-error 
//Argument of type 'string | undefined' is not assignable to parameter of type 'string | number | symbol'.
//   Type 'undefined' is not assignable to type 'string | number | symbol'
//
// but it 'works', because Javascript coerces obj[undefined] to obj['undefined']
// y = 'HELLO' | 'WORLD' | 'MISSING'

const y = tableLookup(data.value)

// after this PR - no type error; y = 'HELLO'|'WORLD'|'MISSING'


const nofallbackLookup = lookup(TABLE);
const z = nofallbackLookup(data.value)
// z = 'HELLO' | 'WORLD' | undefined

```

## ✅ Pull Request Checklist

- [NA] Included link to corresponding [GitHub Issue](https://github.com/gohypergiant/standard-toolkit/issues).
~~- [ ] Added/updated visual regression tests for this change (for bug fixes / features).~~
~~- [ ] Added/updated documentation (for bug fixes / features)~~
~~- [ ] Filled out test instructions.~~
~~- [ ] Added changeset (for bug fixes / features).~~

## 📝 Test Instructions

Run core package tests

- [ ] Yes
- [X] No

## 🤖 AI Usage

100% human

### DIFF
--- a/packages/core/src/utility/lookup/index.ts
+++ b/packages/core/src/utility/lookup/index.ts
@@ -39,17 +39,15 @@ import { identity } from '../../combinators/identity';
  * colorLookup(data.value);
  * ```
  */
-export const lookup =
-  <
-    A extends Record<string | number | symbol, unknown>,
-    // biome-ignore lint/suspicious/noExplicitAny: This is intended
-    B extends (...args: any[]) => any,
-  >(
-    obj: A,
-    def?: B,
-  ) =>
-  <C extends keyof A>(prop: string | number | symbol): A[C] => {
-    const fn = def ?? identity;
-
-    return fn(obj[prop]);
-  };
+export const lookup = <
+  A extends Record<PropertyKey, unknown>,
+  // biome-ignore lint/suspicious/noExplicitAny: This is intended
+  B extends (...args: any[]) => any,
+>(
+  obj: A,
+  def?: B,
+) => {
+  const fn = def ?? identity;
+  return <C extends keyof A>(prop: PropertyKey | null | undefined): A[C] =>
+    fn(prop == null ? undefined : obj[prop]);
+};
--- a/packages/core/src/utility/lookup/index.test.ts
+++ b/packages/core/src/utility/lookup/index.test.ts
@@ -56,8 +56,22 @@ describe('tableLookup', () => {
   it('should use identity for undefined default', () => {
     const actualOne = undefLookup('UNKNOWN');
     const actualTwo = undefLookup('BAR');
+    const actualThree = undefLookup('TEST');
 
     expect(actualOne).not.toBeDefined();
     expect(actualTwo).toEqual(colorTable.BAR);
+    expect(actualThree).toEqual(null);
+  });
+
+  it('should use default for for a null or undefined props', () => {
+    const actualOne = colorLookup(undefined);
+
+    expect(actualOne).toEqual(defaultColor);
+  });
+
+  it('should use identity  for for a null or undefined props', () => {
+    const actualOne = undefLookup(undefined);
+
+    expect(actualOne).toEqual(undefined);
   });
 });


packages/core/src/utility/lookup/index.ts
```diff
@@ -39,17 +39,15 @@ import { identity } from '../../combinators/identity';
...
-  };
+export const lookup = <
+  A extends Record<PropertyKey, unknown>,
+  // biome-ignore lint/suspicious/noExplicitAny: This is intended
+  B extends (...args: any[]) => any,
+>(
+  obj: A,
+  def?: B,
+) => {
+  const fn = def ?? identity;
```
devindavies, line 50: hoisting this makes sense to eliminate per-call nullish coalescing

packages/core/src/utility/lookup/index.ts
```diff
@@ -39,17 +39,15 @@ import { identity } from '../../combinators/identity';
...
+  return <C extends keyof A>(prop: PropertyKey | null | undefined): A[C] =>
+    fn(prop == null ? undefined : obj[prop]);
```
devindavies, line 52: I'm concerned about this branching logic in a hot path, and the possibility of killing v8 (runtime) optimizations. Maybe we have a separate lookup function that allows nullish values?

jemccarthy13: Hm. 
Other option:

```
//@ts-expect-error JS will coerce undefined and it works
fn(obj[prop])
```

belsrc:
That is why there is an `orDefault` function. Several of them in Jerico codebase (pre and post curried). Function composition.

```ts
const colorLookup = lookup(COLOR_MAP, orDefault(COLOR_MAP.PENDING));
```

Check `packages/utils/src/conditionals/or-defaults.ts` for how to curry them re: your `x => x ?? 'MISSING'` line in the OP

jemccarthy13:
> That is why there is an `orDefault` function. Several of them in Jerico codebase (pre and post curried). Function composition.
> 
> ```ts
> const colorLookup = lookup(COLOR_MAP, orDefault(COLOR_MAP.PENDING));
> ```
> 
> Check `packages/utils/src/conditionals/or-defaults.ts` for how to curry them re: your `x => x ?? 'MISSING'` line in the OP

orDefault doesn't solve the type error

the problem is colorLookup(undefined)

[or colorLookup(data) where data is nullable]

colorLookup = lookup(COLOR_TABLE, orDefault('red')) // colorLookup(undefined) should return [255,0,0,255] 
colorLookup =lookup(COLOR_TABLE) // colorLookup(undefined) should return undefined
// but in both cases, colorLookup(undefined) is a type error

in the case of data?.value or data.value where value is optional.. both versions of lookup (with or without default)

jemccarthy13:
To tie both threads together:

In JERIC2O (not authored by me):

```
const colorLookup = lookup(ID_ROE_LOOKUP, orDefault(ID_ROE_COLOR_LOOKUP.PENDING));

const deriveColor = (identity?: IDRoe) => {
  // here because colorLookup doesn't accept undefined
  if (!identity){ 
    return ID_ROE_COLOR_LOOKUP.PENDING;
   }

   return [...colorLookup(identity)];
}
```

Is used in an accessor for tracks.. so fairly hot path, and largely driven only by TS complaining about undefined because this signature doesn't support undefined:
`<C extends keyof A>(prop: string | number | symbol): A[C] => {`

I could do 
`colorLookup(identity ?? 'PENDING')` but that defeats the purpose of the orDefault path


------
technically, this is the 'best' of both worlds I can see - no branch, support for undefined / nullable input to the result of a lookup() construction

```
export const lookup = <
  A extends Record<PropertyKey, unknown>,
  // biome-ignore lint/suspicious/noExplicitAny: This is intended
  B extends (...args: any[]) => any,
>(
  obj: A,
  def?: B,
) => {
  const fn = def ?? identity;
  return <C extends keyof A>(prop: PropertyKey | null | undefined): A[C] =>
    // @ts-expect-error JS will coerce undefined to 'undefined' so this is fine
    fn(obj[prop]);
};
```

belsrc:
The long as boring of it and where I am coming from.

This change violates the input-validation rules from the style guide. It adds defensive `null` checking in an internal leaf function. It widens the type to `PropertyKey | null | undefined` without a trust-boundary reason. The rule states: "Validate and sanitize external data at system boundaries." This principle applies in reverse. Validation belongs at boundaries, not in leaf functions.

The problem is that `lookup` is an internal utility, not a public API boundary. This change makes it defend against its own callers. Validation moves down into the leaf function instead of up to where bad data enters the system. If a caller passes `null`, that is a bug in the caller. Every internal function starts defending against bad input. Which, I admit, is a **huge** problem in Jerico. Checks duplicate throughout the call stack. Now we must test "what if prop is `null`?" in a function that should never receive `null`. 

The ideal approach is to have the caller validate at the trust boundary ("Did I get valid user input?"). Leaf functions should trust their callers ("If you call me, prop is valid"). The original signature documents the constraint: `prop` must be a property key type (`string`/`number`/`symbol`), not `null` or `undefined`.

jemccarthy13:
Discussed; not a good change

